PR #5, opened 2016-02-05
Add only one log handler if OMPython.OMCSession() is called multiple …

fhgd:
…times

Otherwise each logging entry appears multiple time on the screen which is annoying if I start an OMCSession from a script.


RexFuzzle:
This is not the correct way of doing this as the logger should be defined outside of the class. See pull request #6

PR #7, opened 2016-04-06
- Sudeep's thesis work OMPython Enhancement

alash325:


OMPython/__init__.py
```diff
@@ -62,6 +62,21 @@
 import pyparsing
 from distutils import spawn
 
+# The following import are added by Sudeep
+import numpy as np
+import pandas as pd
+import csv
+
+import matplotlib.pyplot as plt
```
sjoelund: Is it wise to depend on matplotlib when it causes the compiler to use exponential memory? https://trac.openmodelica.org/OpenModelica/ticket/3764

OMPython/__init__.py
```diff
@@ -62,6 +62,21 @@
 import pyparsing
 from distutils import spawn
 
+# The following import are added by Sudeep
```
bilderbuchi: any non-stdlib import should also be added to setup.py `install_requires`.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+			print file_
+			self.model = self.fileName_[:-3]
+			if(self.fileName_):
+				os.chdir(self.currDir)
+				self.loadingModel(self.fileName, self.modelName, self.lmodel)
+			else:
+				print "Error: File does not exist!!!"
+				return
+
+				self.create_scalarVariableList()
```
bilderbuchi: due to the `return` two lines above, this code is unreachable, so this is probably a bug. did you run a linter over the code?

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+			specError = 'Parser error: Unexpected token near: optimization (IDENT)'
+			if specError in loadfileError:
+				self.requestApi("setCommandLineOptions", '"+g=Optimica"')
+				self.requestApi("loadFile", fName)
+			else:
+				print 'loadFile Error: ', loadfileError
+			return
+			
+		if lmodel is None:
+			pass
```
bilderbuchi: is this `pass` some leftover section placeholder yet-to-be-implemented? if not, why not make this clause `if lmodel:` instead of the weird if None pass + else construct?

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+				print loadmodelError
+				return
+			
+		buildModelError = ''
+		self.getconn.sendExpression("setCommandLineOptions(\"+d=initialization\")")
+		buildModelResult = self.requestApi("buildModel", mName)
+		buildModelError = self.requestApi("getErrorString")
+		if buildModelError:
+			if 'Message: Fault' in buildModelError:
+				print ''
```
bilderbuchi: print an empty string?

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+					start = att.get('start')
+				self.quantitiesList.append(Quantity(name, start, changable, variability, description, causality))
+		return self.quantitiesList
+		
+	def getQuantitiesNames(self):
+		if self.qNamesList == []:
+			for q in self.quantitiesList:
+				self.qNamesList.append(q.name)
+		else:
+			pass
```
bilderbuchi: empty placeholder for stuff-to-come? else, why would you have an empty else clause?

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+		return self.qNamesList
+		
+	def checkAvailability(self, names, chkList, inputFlag = None):
+		isExist = True
+		if inputFlag is not None:
+			if names in chkList:
```
bilderbuchi: why not `if names not in chkList`, without the `pass`?

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+				pass
+			else:
+				isExist = False
+			if(isExist == False):
```
bilderbuchi: actually, you should be able to replace this with `if names not in chkList`, and drop the if statement above.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+				print 'Error!!! ', names , ' does not exist.'
+				return False
+		
+		else:
+			nonExistingList = []
+			for n in names:
+				if n in chkList:
+					pass
```
bilderbuchi: again an empty if/else clause. from now on i'll just assume this is a bad coding pattern. (would be good to learn to avoid doing that, tbh)

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+				else:
+					isExist = False
+					nonExistingList.append(n)
+
+		if(isExist == False):
```
bilderbuchi: `if not isExist` would be the pythonic way to say this.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+					self.cValuesList.append(l.start)
+		return self.cValuesList
+		
+	def getContinuousValues(self, contiName=None):
+		if contiName is None:
+			if self.cValuesList == []:
+				for l in self.quantitiesList:
+					if(l.variability == "continuous"):
+						str_ = l.start
+						if str_ is None:
```
bilderbuchi: watch out when doing boolean comparison! `is` compares object identity, not content, so it's almost always more appropriate to use `==`, except when comparing to `None`.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+							nameList.append(q.name)
+							startList.append(q.start)
+							changableList.append(q.changable)
+							variabilityList.append(q.variability)
+							descriptionList.append(q.description)
+
+				dic = {'Name' : nameList, 'Value' : startList, 'Changable' : changableList, 'Variability' : variabilityList, 'Description' : descriptionList}	
+				frame = pd.DataFrame(dic)
+			except Exception as e:
+				return e
```
bilderbuchi: probably at least a `print e` should be added here, instead of silently swallowing any exception.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+				return None
+				
+	def getParameterValues(self, paraName = None):
+		if paraName is None:					
+			if self.pValuesList == []:
+				for l in self.quantitiesList:
+					if(l.variability == "parameter"):
+						str_ = l.start
+						if ((str_ is None) or (str_ == 'true' or str_ == 'True' or str_ ==  'False' or str_ == 'false')):
+							if (str_ == 'ture' or str_ == 'True'):
```
bilderbuchi: `ture` seems to be a typo.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+		for v in varList:
+			if v == 'time':
+				continue
+			isExist = False
+			for l in self.quantitiesList:
+				if v == l.name:
+					isExist = True
+				else:
+					continue
+			if(isExist == False):
```
bilderbuchi: afaict, the `for` and `isExist` can be dropped, and the `if` be replaced with
`if v not in [l.name for l in self.quantitiesList]:`

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+			exp = "readSimulationResult(\"" + resFile + '",{' + variables + "})"
+			res = self.getconn.sendExpression(exp)
+			npRes = np.array(res)
+			exp2 = "closeSimulationResultFile()"
+			self.getconn.sendExpression(exp2)
+			return npRes
+		else:
+			return "Error: mat file does not exist"
+					
+	def plot_(self, x_axis, y_axis): #Not complete yet
```
bilderbuchi: better mark this with a `#TODO: ` comment, so that IDEs can pick this up.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+					pass
+				elif isinstance(v,str):
+					return errMsgStr
+				else:
+					return errMsgFormat
+			variability = "variability"
+			attrValue = "continuous"
+			self.setValue(names, values, variability, attrValue, self.cNamesList, self.cValuesList)
+		except Exception as e:
+			return e
```
bilderbuchi: swallowing all exceptions silently is typically a bad pattern, but I'll stop remarking on this now.

OMPython/__init__.py
```diff
@@ -447,3 +462,727 @@ def getClassNames(self, className=None, recursive=False, qualified=False, sort=F
...
+			if inputsValList == sorted(inputsValList):
+				pass
+			else:
+				return errMsgTime
+			
+			index = self.inputs.index(name)
+			self.inputsVal[index] = inputsValList
+			self.inputFlag = True
+		except Exception as e:
+			if (TypeError):
```
bilderbuchi: this will always be true, as `TypeError` is an exception object.

OMPython/__init__.py
bilderbuchi: probably it can be dropped, as it's only used in a [very thin wrapper](https://github.com/OpenModelica/OMPython/pull/7/files#diff-a1709e22cd56663c5abde760173fd676R953) around plt.plot. Same with the other dependencies, pd is only used a couple of times to return some pd dataframes (maybe the user should do that instead if desired)´, and numpy is used only twice (although one is an interp, so may be difficult to replace).

bilderbuchi:
I had a quick glance over this, but stopped reviewing at about L1030.
This PR seems to use Tabs for indentation, as opposed to the rest of the file (afaict), and basically the rest of Python, might be wise to change this before merging.
Also, 700+ lines of diff, and not a single explanatory comment, or class/method/function docstring, and no tests? I really wouldn't want to maintain this tbh.

PR #8, opened 2016-04-12
Fixed comments from Christoph Buchner and Martin

SudeepLiu:
Fixed comments from Christoph Buchner and Martin

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -62,6 +62,13 @@ import tempfile
 import pyparsing
 from distutils import spawn
 
+# The following import are added by Sudeep
+import numpy as np
+import csv
+
+from copy import deepcopy
+import xml.etree.ElementTree as ET
+
 if sys.platform == 'darwin':
     # On Mac let's assume omc is installed here and there might be a broken omniORB installed in a bad place
     sys.path.append('/opt/local/lib/python2.7/site-packages/')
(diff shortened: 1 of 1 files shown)


bilderbuchi:
it is typically more useful to add this to the already existing PR #7 as an additional commit. this way discussion/remarks get preserved, and fixed section are more readily apparent.

bilderbuchi:
also, regarding the comments: if all you do is repeat the function name, this it typically not very useful. It's better to (also) specify details about input arguments, returned objects, potential exceptions.
Additionally, please look up [python docstrings](https://www.python.org/dev/peps/pep-0257/) and a [pretty useful convention for functions](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html) if you add these - the comments you added right now are not parsed by Python and its IDEs, so is not really useful. I would understand if you don't bother, as the rest of the codebase, from a quick glance, is also not very much documented.

As a simple example, I used one of your member functions:
```
   #to get values of continuous quantities name
   def getContinuousValues(self, contiName=None):
```
better would be,e.g.,
```
  def getContinuousValues(self, contiName=None):
  """Return values of continuous quantities.

  If no quantity is given, XXX happens.
  
  Args:
    contiName (Optional[string]): Name of the continuous quantity. Defaults to None.
  Returns:
    list: Values of the continuous quantity queried.
  """
```

bilderbuchi:
also, maybe, before investing this work, coordinate with @sjoelund if he intends to merge the (improved) work.

alash325:
Import FMU is not working in your updated code i.e convertFmu2Mo. I think you removed the following function  def requestApi__(self, apiName, entity ).

SudeepLiu:
Thanks you Alachew for reminding. I forgot that i have been calling that
function. Thats why i told to test by yourself. Well do more tests n let me
know abt more bug.

On Thursday, 14 April 2016, Alachew Mengist <notifications@github.com>
wrote:

> Import FMU is not working in your updated code i.e convertFmu2Mo. I think
> you removed the following function def requestApi__(self, apiName, entity
> ).
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/OpenModelica/OMPython/pull/8#issuecomment-209688351>
>


-- 
With Regards,
Sudeep Bajracharya

PR #9, opened 2016-04-29
add flag parsed=true for sendExpression

arun3688:
Pyparsing not giving error messages in case of errors in code, just throws exception
eg: omc.sendExpression("model test end test")


adeas31:
This is wrong. Just catch the exception and return the error.

PR #10, opened 2016-04-29
catch Pyparsing exceptions

arun3688:


OMPython/__init__.py
```diff
@@ -238,8 +239,11 @@ def sendExpression(self, command):
             self._omc = None
             return result
           else:
-            answer = OMTypedParser.parseString(result)
-            return answer
+            try:
+              answer = OMTypedParser.parseString(result)
+              return answer
+            except:
```
bilderbuchi: 1. It's typically bad python style to catch all exceptions without qualification, you should probably limit this to PyPArsingExceptions or whatever the type of the exceptions you expect are.
2. In the case of the exception, shouldn't you return some info about the exception, e.g. its message? You just return the string you wanted to parse in the first place.

arun3688: If you look into the code, the results of the compiler is returned not the
string we wanted to parse, because the Pyparsing does not give proper error
messages in case there are errors in the code

regards
arun

On Fri, Apr 29, 2016 at 2:37 PM, Christoph Buchner <notifications@github.com

bilderbuchi: well, this is then just confusing nomenclature. this code snippet tries to parse a string called `result`. Let's consider this an **input** of this snippet, or "the string you want to parse" by passing it to a function called `parseString`. What you want to return is an `answer`, let's consider this an **output** of this snippet. 

Now, if `parseString` throws an exception, instead of reporting or returning the exception message, you return `result`, i.e. the **input** of your snippet. You drop all notice of an error having occured, and return an unparsed input, instead of a parsed output.

Btw, I don't know how familiar you are with python, but throwing exception is typically considered "proper error reporting", so the exceptions should contain valuable error information.

arun3688: i think you are not getting what i am saying, The result is output message obtained from the complier, (it can be both correct output and also error message). The Pyparsing in few cases does not give error messages, i hope you are using OMPython try this 
omc.sendExpression("model a end a;")
output- (a,)
omc.sendExpression("model a end a2;")
The output should be a proper error message returned from the compiler and not just the pyparsing exception.

bilderbuchi: Ah, so you are saying that if parseString throws an exception, you assume it's because `result` contains only a compiler error message, and not because of otherwise malformed compiler output, so you return the compiler error message instead of the parsed `answer`, right?

Unfortunately, I'm on Python 3, so not using OMPython (yet).

adeas31: I agree that we should catch PyPArsingExceptions instead of all exceptions and should print its message.

Returning `result` is just fine. Basically from this function we want to return the result of the OMC API call done via `sendExpression`. `OMTypedParser.parseString` is used to format the result in a proper python type instead of just returning string type. So if `parseString` is successful return the formatted result i.e., `answer` otherwise just return the `result`.

sjoelund: No, returning `result` is not fine. How do you distinguish between a function returning a string and a function giving an error? If you wanted just the result, just do a call that does not expect the function to be possible to parse. Or add another OMPython call that tries to parse the string and otherwise always returns the result.

arun3688: what about adding a parsed FLag in send Expression , so that when i want only unparsed result i can specify the flag to be false , like in this commit https://github.com/arun3688/OMPython/commit/ee3287a4b7da62c3cf010e9dd41ec9bab424e5c2

sjoelund: Just use a different python call for parsed=False. It should also fail when failing to parse, no questions asked.

bilderbuchi:
this PR has conflicts now which need to be solved before it becomes mergeable.

adeas31:
@arun3688 can you please fix the conflicts and rebase this.

syntron:
Isn't this PR included in PR #270 or even earlier? handling of the different parsers was updated and improved since this PR ...

PR #14, opened 2016-07-12
Streamline typeCheck and add initial test suite

alchemyst:
I've started by making typeCheck easier to change by changing the order and contents of the types variable rather than redoing logic. In the process I've also started a small test suite to make sure I don't mess anything up with the edits.

I am a bit puzzled by what kind of thing will raise ValueError when passed to `str` in Python, so the final tests are empty right now.

### DIFF
--- a/OMPython/OMParser/__init__.py
+++ b/OMPython/OMParser/__init__.py
@@ -41,36 +41,28 @@ next_set_list = []
 next_set = []
 next_set.append('')
 
+def bool_from_string(string):
+    if string in {'true', 'True', 'TRUE'}:
+        return True
+    elif string in {'false', 'False', 'FALSE'}:
+        return False
+    else:
+        raise ValueError
+
 def typeCheck(string):
-    if "\n" in string:
-        new_line_char = string[-1]
-        if new_line_char == "\n":
-            string = string.replace(string[-1],'').strip()
+    """Attempt conversion of string to a usable value"""
+    types = [bool_from_string, int, float, long, dict, str]
 
-    if string == "true" or string == "True" or string == "TRUE":
-        string = True
-        return string
-    elif string == "false" or string == "False" or string == "FALSE":
-        string = False
-        return string
+    string = string.strip()
 
-    try:
-        string = int(string)
-    except ValueError:
+    for t in types:
         try:
-            string = float(string)
+            return t(string)
         except ValueError:
-            try:
-                string = long(string)
-            except ValueError:
-                try:
-                    string = dict(string)
-                except ValueError:
-                    try:
-                        string = str(string)
-                    except ValueError:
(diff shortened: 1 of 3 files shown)


OMPython/OMParser/__init__.py
```diff
@@ -41,36 +41,28 @@
 next_set = []
 next_set.append('')
 
+def bool_from_string(string):
+    if string in {'true', 'True', 'TRUE'}:
```
bilderbuchi: That would be simpler with `if string.lower() == 'true':

OMPython/OMParser/__init__.py
```diff
@@ -41,36 +41,28 @@
...
+        return True
+    elif string in {'false', 'False', 'FALSE'}:
```
bilderbuchi: see above

tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+from OMPython import OMParser
```
bilderbuchi: why are you importing here and in the test_*.py files? is this a unittest quirk? (I'm mainly familiar with pytest)

bilderbuchi: btw, :+1: on including tests in your code!

OMPython/OMParser/__init__.py
alchemyst: I also thought about that, but that would actually do something different from the current implementation, which only checks for those particular strings. This change was meant to be a pure refactor.

tests/__init__.py
alchemyst: I definitely need to import from the `test_*.py` files for the tests to run (using `python -m unittest tests tests.test_OMParser` from the base directory). I'm not very used to working with packages (the current structure seems very baroque to me) but this is what I had to do to get the tests going. I'm open to improvement.

OMPython/OMParser/__init__.py
bilderbuchi: OK, right, this makes sense.

tests/__init__.py
bilderbuchi: Yeah, i was primarily wondering about i.porting it in __init__.py, then not using it.

alchemyst: You're right. I've removed it.

alchemyst:
Is there anything else I can change to have this pull request accepted?

bilderbuchi:
we can ping @sjoelund...

alchemyst:
Any news on this in the new year?

bilderbuchi:
I don't know, I'm just a drive-by potential user. @adeas31 apparently has merge rights, as he recently merged another PR, so maybe he can take a look?

alchemyst:
Thanks @hkiel 😀

PR #22, opened 2017-05-25
Improvements to behaviour of ModelicaSystem class

adbuerger:
I think it would be useful to add these changes to OMPython:

1.) When calling getParameters() and if a parameter is a string that is not true or false and cannot be converted to a float, return the string value instead of throwing an error, since the parameter then probably is a string (e. g. a file path)

2.) When calling getSolutions() and no solution is present yet, the undefined function exit() is called, while I think sys.exit() should be called

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -1166,8 +1166,7 @@ class ModelicaSystem(object):
         res_mat = '_res.mat'
         resFile = "".join([self.modelName, res_mat])
         if (not os.path.exists(resFile)):
-            print ("Error: Result file does not exist")
-            sys.exit()
+            raise IOError("Result file does not exist")
         else:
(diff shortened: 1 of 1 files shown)


sjoelund:
I don't think it should exit the process almost regardless of what happens. Better to raise an error and possibly quit the OM connection if it's fatal.

⁣Sent from BlueMail ​

adbuerger:
@sjoelund Thanks for you remark, you are right, I changed it accordingly. I only used sys.exit() since I thought it was the function intended to be called with the previously undefined function exit().

syntron:
@adeas31 this is a quite old issue - (2) is solved and I think (1) should also work (not tested)

adeas31:
OK then I will close this.

Issue #1, opened 2015-05-05
Python 3 support?


bilderbuchi:
It seems OmniORB is the holdup keeping Python support to version 2. As I was digging around, I found that apparently the dev version/upcoming version offer Python 3 support, judging from [some commit messages since 07/2014](http://sourceforge.net/p/omniorb/svn/commit_browser) and even some parts of the current version [support Py3](http://www.omniorb-support.com/pipermail/omniorb-list/2014-April/031550.html), so I thought I'd leave that here.

bilderbuchi:
I just [saw](https://trac.openmodelica.org/OpenModelica/changeset/24682) that OMEdit got rid of Corba - does that mean that Corba in OM is on its way out, or is this restricted to OMEdit? If the former that would probably help with a Python3 transition.

sjoelund:
The Python interface could indeed be changed to use dynamic loading as well. However, it is not as simple as for the C++ interface since that interface can access the macros used to start the garbage collector, create MetaModelica strings, and so on. For Python, all such interactions would probably need to be performed on a separate thread. Which means coding a Python C wrapper instead of CORBA...

bilderbuchi:
So, omniORB/omniORBpy 4.2.1 is now out, [having full Python3 support](https://sourceforge.net/p/omniorb/svn/HEAD/tree/trunk/omniORBpy/ReleaseNotes.txt) (Python 3.2+/2.5+). :-)

sjoelund:
4.2.1: This release has the following known limitations: Only Python 2.x is supported.
So their biggest new feature was not removed from the limitations of the release notes :D

bilderbuchi:
yeah, I saw this too, and I concluded that it was just an oversight (probably late-night release preparation ^^).

RexFuzzle:
Any news on Python3 support? Just starting to run into some other issues that it seems python3 might be able to solve but OMPython is holding me back from updating.

bilderbuchi:
not afaik, but OmniORB in fact seems to be py3-compatible, I see py3-related commit messages in [the log](https://sourceforge.net/p/omniorb/svn/commit_browser).

hkiel:
From the build for windows it seems Python3 really **is** supported:
https://sourceforge.net/projects/omniorb/files/omniORBpy/omniORBpy-4.2.1/omniORBpy-4.2.1-win64-python35.zip

hkiel:
[...]

PR #23, opened 2017-06-15
Added ZeroMQ client for OMC

adeas31:
Implemented a new class OMCSessionZMQ. OMCSession still exits. Both OMCSession and OMCSessionZMQ inherits from OMCSessionBase.
ModelicaSystem class has a new parameter useCorba which is False by default. So ModelicaSystem by default always use OMCSessionZMQ. If you want to use OMCSession then set `useCorba=True`.

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -80,7 +80,7 @@ if sys.platform == 'darwin':
 from OMPython import OMTypedParser, OMParser
 
 # Logger Defined
-logger = logging.getLogger('OMCSession')
+logger = logging.getLogger('OMPython')
 logger.setLevel(logging.DEBUG)
 # create console handler with a higher log level
 logger_console_handler = logging.StreamHandler()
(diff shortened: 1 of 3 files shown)


hkiel:
@adeas31 Are you going to implement ZMQ for OMNotebook as well?

adeas31:
@hkiel why do you need ZMQ for OMNotebook?
OMNotebook directly loads the OpenModelicaCompiler binary and communicate via the C interface.

thorade:
Maybe he wanted to ask about the Jupyter kernel, not OMNotebook?
https://github.com/OpenModelica/jupyter-openmodelica

adeas31:
Well openmodelica jupyter kernal uses OMPython.
However, it is using `OMCSession` perhaps we can update it to use `OMCSessionZMQ`.

bilderbuchi:
Great to see this! :+1:

dietmarw:
So does the jupyter kernel now already use OMCSessionZMQ?

dietmarw:
Also it looks like the documentation was forgotten as to what packages OMPython now needs to run. Could this be updated?

arun3688:
No the jupyter kernel is not using ZMQ, i will change it by next week

regards
arun

thorade:
I just tried to install it (on Travis), but installation fails with error message 
`FileNotFoundError: [Errno 2] No such file or directory: 'omniidl'`
Is that still required? There are no pypi packages for omniidl.

hkiel:
I was referring to OMMNotebook and OMShell as these currently use corba for communication to omc

adeas31:
OMNotebook does not use CORBA. I updated it to directly link the OpenModelica binary long ago.
OMShell still uses CORBA and is fine with it.

adeas31:
@dietmarw I updated the version number.
@thorade It should be fixed in #24. Try again. Note that you also need the latest omc with ZeroMQ so that `omc -d=interactiveZMQ` is possible.

adeas31:
> I dislike you naming the flag -d=interactiveZMQ though... Should have been changed to --interactive=none|tcp|zmq|corba or something while we were fixing things :)

Not really. That change wouldn't be backward compatible. We do have a lot of people using CORBA interface via `-d=interactiveCorba`.

dietmarw:
Why not implement it in the proposed cleaner way like @sjoelund suggested and leave the alias `-d=interactiveCorba` -> `--interactive=corba` with a deprecation warning for backward compatibility?

pierre-haessig:
I've seen the README updated to mention PyZMQ as a dependency. What is not clear to me is whether now omniORB is still a *hard dependency* or if it is *optional*.

As a related question, is OMPython still compatible with Python <=2.7 only ?

dietmarw:
The whole README is currently not usable when it comes to installation. The path given don't exist on linux since OMPython is not packaged with the Linux version. There is still the unnecessary git dependency in the instructions.

So the README should have clean instructions for Windows and Linux separate which are tested. Otherwise new users will just be frightened away from OMPython ;-)

thorade:
@pierre-haessig Python 3 compatibility is also discussed in #1.

adeas31:
>> I dislike you naming the flag -d=interactiveZMQ though... Should have been changed to --interactive=none|tcp|zmq|corba or something while we were fixing things :)

> Not really. That change wouldn't be backward compatible. We do have a lot of people using CORBA interface via -d=interactiveCorba.

@sjoelund @dietmarw See https://github.com/OpenModelica/OMCompiler/pull/1687

pierre-haessig:
As a side question, is ZMQ available from within a Modelica simulation or is it specific to OMC ?

adeas31:
> As a side question, is ZMQ available from within a Modelica simulation or is it specific to OMC ?

It is specific to OMC.

arun3688:
@dietmarw , The Jupyter kernel for OpenModelica is now changed to use OMCSessionZMQ

dietmarw:
@arun3688 Great. With "now" you mean today so it will be in the next nightly?

arun3688:
@dietmarw , the jupyter kernel is not part of Nightly builds, but OMPython comes under nightly builds, But anyhow the jupyter openmodelica kernel should run in python 3 as well as

dietmarw:
OK, I wasn't sure if the kernel changes would also need some omc code changes.

PR #30, opened 2017-06-16
Send and receive strings from zmq

adeas31:


thorade:
That makes it work for me on WinPython 3.4 64 bit
Awesome!

PR #31, opened 2017-06-16
ignore some more files

thorade:
ignore .pyc and .pyo and .pyd files
ignore .bak files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /build/
 /dist/
 /tmp/
+*.py[cod]
+*.bak


bilderbuchi:
Take care, if some of those newly ignored files are already tracked in the repo, they are not ignored, and you have to remove them manually (and commit) once.

thorade:
OK, thanks. As far as I can see no .pyc/.bak files are in the repo, but they were annoying me when using OMPython today.

bilderbuchi:
Sure, I can imagine. A good gitgnore is pretty important to not clutter your repo...

PR #32, opened 2017-06-16
Use the new --interactive flag.

adeas31:


hkiel:
@adeas31 please do "rebase and merge" when merging a PR. Maybe we can disable the other two options like in OMCompiler repo!?!

sjoelund:
@adeas31 OMPython needs to be able to use the old flag (at least as an option) because we use OMPython to run old OMC versions (for example for the library testing script).

PR #37, opened 2017-06-20
autopep8 --max-line-length=999

### LINKED ISSUE #35: pep8
Mostly whitespace changes, but it might increase readability to follow pep8
https://www.python.org/dev/peps/pep-0008/

To clean the code atuomatically, run
```
autopep8 --in-place --recursive --max-line-length=120 OMPython
```

Using a line length of 120 is recommended here:
https://stackoverflow.com/questions/22207920/what-is-githubs-character-limit-or-line-length-for-viewing-files-on-github
but other values, e.g. 999, also have advantages.
Do not specify line length at all to use the pep8 default
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

thorade:
this closes #35
exact command used:
```
autopep8 --in-place --recursive --max-line-length=999 OMPython
```

That uses long lines to avoid addition of line breaks. Shorter lines would be nicer in diffs, but automatic line breaks are not optimal in all cases and require manual tweaking.

To see the same commit without whitepsace, append `?w=1`to the commit URL:
https://github.com/OpenModelica/OMPython/commit/ec9f11cb35a1be8f6eb3197999ac7b9e4aed8e8b?w=1
That diff is much smaller and leaves mostly added/removed blank lines.

If those changes are also undesired, some or all E3 rules can be excluded
http://pep8.readthedocs.io/en/stable/intro.html#error-codes

### DIFF
--- a/OMPython/OMParser/__init__.py
+++ b/OMPython/OMParser/__init__.py
@@ -41,6 +41,7 @@ next_set_list = []
 next_set = []
 next_set.append('')
 
+
 def bool_from_string(string):
     if string in {'true', 'True', 'TRUE'}:
         return True
@@ -48,6 +49,7 @@ def bool_from_string(string):
         return False
     else:
         raise ValueError
+
 
 def typeCheck(string):
     """Attempt conversion of string to a usable value"""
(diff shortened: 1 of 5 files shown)


adeas31:
@arun3688 Can you test that this doesn't break anything?

arun3688:
@adeas31 I tested this pull request, i tested some basic APIS it works it does not break anything

thorade:
Thanks for testing & merging.
From now on one can re-run autopep8 again and again and it should give very small diffs only.

PR #38, opened 2017-06-21
more aggressive autopep8

thorade:
addendum to #35 
increase autopep8 aggressiveness by using
```
autopep8 --aggressive --aggressive --aggressive --in-place --recursive --max-line-length=999 OMPython
```


bilderbuchi:
Huh, i didn't realise that autopep8 could do non-whitespace changes, like the switch from type() to is instance().

PR #24, opened 2017-06-15
Don't try to generate OMPythonIDL file omniidl is not available.

adeas31:
Updated installation notes in README.md

PR #41, opened 2017-06-21
remove sys.platform check

thorade:
related: #24 
only check whether omniidl can be imported, independent of OS

This is just a suggestion: If people follow the installation instructions in the readme, the `import omniidl` will succeed on Win32, ~~so the result is the same as before~~.
People who on Windows use the Linux pip command from the Readme will at least get a working ZMQ version this way.

Edit:
I just noticed that on win32 the function generateIDL() did not get called. Does it get called elsewhere?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,16 +39,13 @@ def generateIDL():
     print("Generated OMPythonIDL files")
 
 
-if sys.platform != 'win32':
-    try:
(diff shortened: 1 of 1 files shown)


thorade:
I just noticed that on win32 the function `generateIDL()` did not get called. Does it get called elsewhere?

PR #46, opened 2017-07-19
Py 2/3 compatible idioms

thorade:
long is gone in Py3
https://docs.python.org/3.0/whatsnew/3.0.html#integers
use int instead, as explained on
http://python-future.org/compatible_idioms.html#long-integers

Does the line 
`from builtins import int`
have to be added to all other files as well?

@alchemyst You wrote that one test:
https://github.com/OpenModelica/OMPython/blame/master/tests/test_OMParser.py
Could you also review this PR, or the test part?

### DIFF
--- a/OMPython/OMParser/__init__.py
+++ b/OMPython/OMParser/__init__.py
@@ -32,9 +32,12 @@
  Version: 1.0
 """
 
-import sys
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from builtins import int, range
 
-from builtins import int
+import sys
 
 result = dict()
 
--- a/OMPython/OMParser/__init__.py
+++ b/OMPython/OMParser/__init__.py
@@ -34,6 +34,8 @@
 
 import sys
 
+from builtins import int
+
 result = dict()
 
 inner_sets = []
@@ -54,7 +56,7 @@ def bool_from_string(string):
 
 def typeCheck(string):
     """Attempt conversion of string to a usable value"""
-    types = [bool_from_string, int, float, long, dict, str]
+    types = [bool_from_string, int, float, dict, str]
 
     string = string.strip()
 
--- a/OMPython/OMTypedParser.py
+++ b/OMPython/OMTypedParser.py
@@ -35,7 +35,10 @@ __license__ = """
 __status__ = "Prototype"
 __maintainer__ = "https://openmodelica.org"
 
-import sys
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from builtins import int, range
 
 from pyparsing import (
     Combine,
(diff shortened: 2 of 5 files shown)


thorade:
Added some more commits, now it should cover almost everything for #1.

bilderbuchi:
👍 🎉

thorade:
@bilderbuchi Would you be able to help with testing? On 2&3 ideally.

bilderbuchi:
Yes I think so, but probably only in a couple of weeks, pretty swamped right now. I can run the tests, at least, as I haven't yet used OMPython (due to no py3 compatibility) in any meaningful way.

thorade:
Pling: Any comments on this? It seems to work for me, but I do not use OMPython very often.

thorade:
@adeas31 Any comments?

@bilderbuchi and @dietmarw and @grantstephens and @forgi86 and everybody that commented in #1 
Does OMPython work for you already with Python3? Does this PR fix it? Anything missing or broken?

thorade:
Here is the script that I ran as a test. 
Before the PR it failed on Python3, after the PR it works on Py2 and Py3
```python

omc = OMCSessionZMQ()

cmds = [
    "getVersion()",
    "loadModel(Modelica)",
    "getClassNames()",
    "simulate(Modelica.Electrical.Analog.Examples.HeatingResistor)",
    ]

for cmd in cmds:
    answer = omc.sendExpression(cmd)
    print("\n{}:\n{}".format(cmd, answer))
```

adeas31:
This looks good and works both with Py2 and Py3.
Should I merge it or do you want to push something more?

thorade:
I pushed one commit yesterday that I had forgotten to push. 
This PR as it is removes all Py2-only code, and I believe it does not break anything, so yes, please merge it.

But some more work will be necessary in the future, because there are other models that still do not work. Plus, it would be nice if you could also update the omniORB python code to the latest version.

bilderbuchi:
@adeas31 thank you for testing! unfortunately, as my current projects, not modelica-related, keep me too busy I have not found the time to test, sorry for that. :-/

adeas31:
@bilderbuchi no problem.
@thorade why do you need to update omniORB? For Py3 compatibility?

thorade:
I read your comment in #1 and updating the omniORB code is not really needed, I just thought it would be easy to update it. And yes, Py3 compatibility is the only benefit that I know of, but I think I prefer ZMQ anyway.

adeas31:
I remember now why updating omniORB is a problem.
We have omniORBpy 4.2.0 and for all newer versions they only provide 64-bit binaries which means we have to compile by our own for 32-bit which is a real pain. Also it depends on which python version you have used to build it.

Issue #39, opened 2017-06-21
pylint


thorade:
As mentioned by @dietmarw in #1 you should run a code analysis with [pylint ](https://www.pylint.org/)or some other tool.
If you happen to use Spyder as your editor, this can be done (for the currently active file) by just pressing F8.
At least the errors should be fixed (or looked at), some of them are due to Py2 vs Py3, some of them are due to omniORB not being installed on my Windows machine, but some are real errors that need a fix.

![spyder_codeanalysis](https://user-images.githubusercontent.com/314571/27372562-5b4eed0a-5665-11e7-8795-8fba1988fd85.png)

thorade:
@adeas31 Which one of the two getPackages functions is correct?
https://github.com/OpenModelica/OMPython/blob/master/OMPython/__init__.py#L264-L268
One of the two should be removed.

adeas31:
Actually both are correct but we can merge them as one,
```
def getPackages(self, className="AllLoadedClasses"):
        return self.ask('getPackages', className)
```

Will you make a PR?

thorade:
pylint reports a couple of time:
E1101: ex has no 'message' member
https://stackoverflow.com/a/33239954/874701

thorade:
Closing for now, would be good to re-run the test on a regular basis.

PR #50, opened 2017-10-17
these exceptions do not have a 'message' member

thorade:
but I am not 100% sure this version is what you wanted....
What does the logger do? Or are you maybe adding a message to the exceptions elsewhere?

For a discussion, also see
https://stackoverflow.com/a/33239954/874701
https://github.com/OpenModelica/OMPython/issues/39#issuecomment-316046301


bilderbuchi:
you could even use the `exc_info=True` [kwarg](https://docs.python.org/3.6/library/logging.html?highlight=logging#logging.Logger.debug) of `logging.(error|warning|debug)` calls for even easier handling.

thorade:
Sorry, I do not know how to use that. Could you just update the PR? I added you as a contributor to my fork, feel free to work in the issue-branch.

thorade:
OK, after reading the linked docs: 
You suggest to change
```python
logger.warning('OMPython error: {0}'.format(ex))
```
to 
```python
logger.warning('OMPython error: {0}'.format(ex), exc_info=True)
```
correct?
Better here in this PR or for ALL logging, in a second PR?

bilderbuchi:
Not exactly. I'll send you a PR/some code later (in the evening) to look at. Not for all of logging, only when you are catching exceptions.

bilderbuchi:
PR sent. I meant like `logger.warning('OMPython error:', exc_info=True)`, but on second thought it's maybe not so smart after all.

thorade:
FYI: Here is another gist with a CLA text:
https://gist.github.com/thorade/5135bbc4f3d5a6cc81272160966910e8
Maybe you also want to include some sentences as in that gist!?

sjoelund:
Good idea. But now you need to sign the CLA again :)

thorade:
Signed it again and noticed that the check mark was green all the time!?

sjoelund:
Yes, I think github doesn't re-run checks once you pass (unless you add a new commit, etc).

thorade:
https://cla-assistant.io/ also has a button with three dots on the right, for every linked repo, there you can trigger a new check it seems.

bilderbuchi:
Wow I did not think one would have to sign a CLA like this, which says, in essence, "I am allowed to contribute code under the project's licence".. :-/
Also a couple remarks:
* The licence text from the CLA linked from this PR, and the one found in the OMPython source code, is not identical, see [this passage](https://github.com/OpenModelica/OMPython/blob/264c572b6ea290eaa529b269b0bbc9b0e7b7204a/OMPython/__init__.py#L72-L73) which is missing the BSD part in the CLA text. You might want to fix that.
* I really think (personal feeling) that the license should be distributed with the source, e.g. in a LICENCE.txt file, and not via some 3rd party service where you only see it when making a PR, or via a website which may or may not be available in the future.
* Additionally, the links given in the source code saying where the licence you agree to can be "found", don't obviously contain the license - eg. why not link to https://www.openmodelica.org/useresresources/license instead? Would of course be solved by shipping the licence, too.

thorade:
@bilderbuchi The license will be shipped with the code, that is not changed.

The CLA is saying: The contributor grants Openmodelica the rights to distribute their code (including the small contribution) under the license mentioned in the CLA. In this case here: I do allow OM to ship OMPython under the OM license.

Regarding the rest: Of course the license text should be identical in all places!

bilderbuchi:
> I do allow OM to ship OMPython under the OM license.

Exactly. But isn't that clear/obvious already when every relevant source file has a header with the license in question? Meaning, has there ever been a court case or whatnot where people said "I cloned a repo which mentioned $license in the readme, on the repo page, and in $license headers in the files I've changed and contributed back. Who could have thought that my contributions fall under $license?" Or is there another reason for such an additional process which seems quite superfluous? I never had to do that before for any OSS software I contributed to.

> @bilderbuchi The license will be shipped with the code, that is not changed.

Well, if I install OMPython using `pip install https://github.com/OpenModelica/OMPython.git` (or whatever the exact command is, then I don't get a copy of the license. Also, there's none in my local clone of the repo. 
All of which is fair game, I think, as the license header point to an URL, it's just not very convenient, especially considering that many other OSS projects just go with a LICENSE file it might as well be standard behaviour, and considering that the license cannot be found under that URL, but you have to dig more to find the page with the file.

bilderbuchi:
> Regarding the rest: Of course the license text should be identical in all places!

So the BSD-new option will be added to the CLA? Otherwise, you'd remove the BSD-new license it previously had, which I would wildly guess needs the agreement of all previous contributors?

sjoelund:
The CLA applies to the entire OpenModelica organization's repositories. I didn't remember we actually have two different versions of the license (the BSD option is the ["runtime" header](https://openmodelica.org/osmc-pl/osmc-pl-1.2-runtime.txt), but the runtime header also licenses code under OSMC-PL license which means we need a copyright license transfer all the same (I am updating the CLA to note that there are two different headers). We have signed contributor agreements on paper from most contributors; but it seems more convenient to let GitHub automate it and make it more clear to contributors what is going on.

The repositories should have a COPYING.md file in them, and 1.1 license versions changed to 1.2, but I guess the OMPython repository was overlooked for some time, and I know in some cases people copy-pasted old code to give new code version 1.1 of the OSMC-PL even though the OSMC board gave the order to convert all license headers several years ago...

bilderbuchi:
Ah, the copying file is there, but it also only contains the header. Thanks for the deeper insight into the whats and whys.

PR #56, opened 2018-02-12
Add a note about further dependencies

lochel:


thorade:
The required dependencies are listed in the setup.py file:
https://github.com/OpenModelica/OMPython/blob/master/setup.py#L66-L71
pip should usually install these dependencies automatically, so the users will usually not have to run the commands you added. Did that not work for you?

Alternatively, it could be helpfull to add a requirements.txt file, it is a simple text file listing the requirements. But then care has to be taken to keep setup.py and requirements. txt in sync.
https://pip.readthedocs.io/en/1.1/requirements.html

lochel:
You are right. I did not run the setup.py, because I am not so familiar with Python.

thorade:
pip should run setup.py for you. pip can also install from a local drive, or from a git repository, or from a zip file.

PR #58, opened 2018-02-15
show what file we are searching for to connect to OMC

adrpo:


adeas31:
Isn't it enough to just print the file path we looked for if we fail to connect.

sjoelund:
@adeas31 I did that in #67 which we hopefully will merge

PR #61, opened 2018-04-11
Fix issue with passing quit() to OMC in subsequent calls of OMCSessionZMQ

ne3x7:
Hello! I ran into a strange problem, found its source and propose a quick fix. Would be very proud to contribute.

The problem occurs when executing multiple subsequent calls to create and close `OMCSessionZMQ`, for example subsequently creating and managing instances of `ModelicaSystem` in a loop.

The issue is that at second (and possibly further) repetitions `sendExpression('quit()')` deadlocks in `self._omc.recv_string()`. The reason for that, I believe, is improper impact of `sendExpression('quit()')` on OMC in the first repetition.

This is a quick fix to allow for desired functionality. The best way to fix this issue is to replace REP/REQ logic in ZMQ to PUSH/PULL logic, which doesn't require responses.


sjoelund:
I cherry-picked and fixed your commit to #67. Perhaps this fix helps with the odd behavior I saw when trying to run OMPython on our Jenkins server.

ne3x7:
Thanks for your response. In order to reproduce the issue, prepare a “model.mo” file with a Class class in it and then run:

```
from OMPython import OMModel

def worker():
    m = OMModel(“model.mo”, “Class”)
    m.simulate()

for _ in range(10):
    worker()
```

The issue I encounter is that `worker()` doesn’t return the second time.

sjoelund:
This seems to work fine on my machine with the old master. But if it is a deadlock scenario it might just be that my machine is too fast or simply not overloaded. I'll try to add this test to the new test suite.

ne3x7:
This might be the case. I run it on a virtual machine with very limited specs.

sjoelund:
OK, I added some tests. They don't deadlock because I have quite many fixes for weird behavior of quitting the OM process, including yours which you say fixed the problem as well.

PR #67, opened 2018-06-05
Add tests for actually calling OpenModelica

sjoelund:
Closes #58 
Closes #61

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 *.sln
 .idea/
 .vs/
+.DS_Store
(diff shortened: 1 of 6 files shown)


sjoelund:
I had quite a lot of problems with OMPython sometimes hanging when trying to quit. This might be related to the problems we have with OMSimulator (@lochel). Do any of you have any better ideas? #61 related perhaps? I think one thing we did wrong was that we just checked the exit status of OMC, but this was actually never updated because we didn't poll or wait on the subprocess.

PR #76, opened 2018-10-04
parse single quoted names correctly

hkiel:
Fixes issue with icon generation of ModelicaCompliance


hkiel:
@sjoelund This fixes another error on OpenModelica_doc target on hudson.
Can you install this new version on hudson, please?

hkiel:
@sjoelund Can you please release this fix as 3.1.2 and update hudson build machine?
This should remove the last crash or OpenModelica_DOC job.

PR #77, opened 2018-10-31
Release v3.1.2

sjoelund:


hkiel:
Is hudson server updated automatically so that OpenModelica_DOC job uses current version?

sjoelund:
No, but I have updated it.

PR #90, opened 2019-05-23
Fix setup.py generateIDL errors

ksmyth:
Fix finding omc by PATH env var: os.path.realpath(None) returns os.getcwd() (at least on Windows)

Fix missing omniidl producing an exception even when errorOnFailure==False


adeas31:
This PR is not valid anymore.

PR #101, opened 2019-09-13
check for omc process status to avoid freezing

arun3688:


JKRT:
@arun3688 Is this related to the issue I discussed with you and @lochel yesterday?

arun3688:
@JKRT , Yes this should fix that issue, ZMQ poller also needs some kind of time out parameter, and i think this should be a more cleaner solution checking for the omc process status

JKRT:
@arun3688 It was an issue here as well then, we should check the same thing in OMJulia and OMatlab I guess

arun3688:
@JKRT Yes i also found the solution for OMJulia and OMMatlab also i will update it sooner

PR #113, opened 2020-02-09
set commandLienOptions in ModelicaSystem

arun3688:
### Purpose:
Allow commandLineOptions to be set in the ModelicaSystem Constructor

### Usage:
```
mod = ModelicaSystem("BouncingBall.mo", "BouncingBall", commandLineOptions="-d=newInst")

```


adeas31:
@arun3688 please update the users guide as well.

arun3688:
sure

PR #123, opened 2020-09-02
Allow to clear OMParser result dictionary

adeas31:


sjoelund (approved):
Approved my my small change

PR #127, opened 2020-10-30
fix ticket #6084

adrpo:


arun3688 (approved):

adrpo:
Thanks @arun3688. I have no idea if the code will work fine, can you send more arguments like that to self.requestApi?

PR #129, opened 2020-11-16
Add Docker support

sjoelund:
For example, use `OMCSessionZMQ(docker="openmodelica/openmodelica:v1.16.0-minimal")`

### DIFF
--- a/.jenkins/python2/Dockerfile
+++ b/.jenkins/python2/Dockerfile
@@ -7,4 +7,4 @@ RUN apt-get update \
   && apt-get update \
   && apt-get install -qy --no-install-recommends omc \
   && rm -rf /var/lib/apt/lists/*
-RUN pip2 install --no-cache pytest
+RUN pip2 install --no-cache pytest psutil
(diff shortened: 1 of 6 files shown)


adeas31:
Gives this error on Windows,

```
def _create_omc_log_file(self, suffix):
        if sys.platform == 'win32':
            self._omc_log_file = open(os.path.join(self._temp_dir, "openmodelica.{0}.{1}.log".format(suffix, self._random_string)), 'w')
        else:
            try:
              self._currentUser = getpass.getuser()
              if not self._currentUser:
                  self._currentUser = "nobody"
            except KeyError:
              # We are running as a uid not existing in the password database... Pretend we are nobody
              self._currentUser = "nobody"
            # this file must be closed in the destructor
            self._omc_log_file = open(os.path.join(self._temp_dir, "openmodelica.{0}.{1}.{2}.log".format(self._currentUser, suffix, self._random_string)), 'w')
```

adeas31 (dismissed):
Add `psutil` to `setup.py`

adeas31:
Maybe add an example with port use.

sjoelund:
I thought to add that to the OM documentation. There are no example commands in this repository, right? I looked for it and didn't find any...

sjoelund:
Or do you mean testcases? I'm a little hesitant since it might randomly fail...

adeas31:
I mean to add a testcase for it.

As that is the only way in Windows so it has to be clearly mentioned in users guide. I also suggest to write it in the README.

adeas31 (approved):

sjoelund:
The README only has `help(OMPython)` :D

Issue #130, opened 2020-11-25
Error in zmq interface


adeas31:
Only happens when you have some error message. Is probably because of my swedish locale.

```
Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from OMPython import OMCSessionZMQ
>>> omc = OMCSessionZMQ()
>>> omc.sendExpression("loadModel(Modelica)")
True
>>> omc.sendExpression("getNthInheritedClass(Modelica.Fluid.Machines.Pump, 1)")
'Modelica.Fluid.Machines.BaseClasses.PartialPump'
>>> omc.sendExpression("getErrorString()")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\OpenModelica\OMWebService\myenv\lib\site-packages\OMPython\__init__.py", line 776, in sendExpression
    result = self._omc.recv_string()
  File "C:\OpenModelica\OMWebService\myenv\lib\site-packages\zmq\sugar\socket.py", line 684, in recv_string
    return self._deserialize(msg, lambda buf: buf.decode(encoding))
  File "C:\OpenModelica\OMWebService\myenv\lib\site-packages\zmq\sugar\socket.py", line 590, in _deserialize
    return load(recvd)
  File "C:\OpenModelica\OMWebService\myenv\lib\site-packages\zmq\sugar\socket.py", line 684, in <lambda>
    return self._deserialize(msg, lambda buf: buf.decode(encoding))
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 108: invalid continuation byte
>>>
```

arun3688:
should this work?  `omc.sendExpression("getErrorString()").encode().decode()`

bilderbuchi:
Is whatever `getErrorString()` returns encoded in utf-8 (and not e.g. this Windows-specific encoding)? 
Is there a "J**ä**vla" in the error ;-)? (0xe4 seems to be "ä")

adeas31:
> should this work? `omc.sendExpression("getErrorString()").encode().decode()`

This didn't work.

sjoelund:
You'll need to fix OMC's gettext to return UTF-8 text instead of Windows-1252 I guess. All strings should be Unicode...

sjoelund:
[...]

PR #131, opened 2020-11-27
Avoid using translations

adeas31:
Workaround for #130

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -555,7 +555,7 @@ class OMCSession(OMCSessionHelper, OMCSessionBase):
         self._timeout = timeout
         self._create_omc_log_file("port")
 
-        self._set_omc_command([serverFlag, "--locale=C", "+c={0}".format(self._random_string)])
+        self._set_omc_command([serverFlag, "+c={0}".format(self._random_string)])
 
         # start up omc executable, which is waiting for the CORBA connection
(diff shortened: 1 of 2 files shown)


OMPython/__init__.py
```diff
@@ -698,6 +698,7 @@ def __init__(self, readonly=False, timeout = 10.00, docker = None, dockerContain
         # set omc executable path and args
         self._set_omc_command([
                                "--interactive=zmq",
+                               "--locale=C",
```
sjoelund, line 701: This won't work with some older OMC versions

adeas31: Are you talking about like 1.12 or later. Anyway people using such old OMC can use older pip install version of OMPython.

bilderbuchi: then that information concerning the minimum OMC version should probably be put somewhere (Readme, setup.py, warning at import time?)

sjoelund: I load 1.9 versions of OMC that are older than this, and they work just fine... For zmq you could put it, but I guess not CORBA

adeas31:
Since the translations are removed from OMC so this issue will not be seen in the future versions anyway. One idea is to close this PR. What do you think @sjoelund? Shall I close or merge this?

casella:
RIP multilingual support. I always switched to English anyway...

PR #132, opened 2021-01-07
Replaced deprecated 'getchildren' method with 'list(elem)' for Python >= 3.2

kzscisoft:
See:
https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren

The `getchildren` method within `xml.etree` was deprecated since Python3.2, as of Python3.9 the method has now been removed This patch applies the suggestion given above.


arun3688:
@kzscisoft can you sign the CLA, so that it can be merged

kzscisoft:
> @kzscisoft can you sign the CLA, so that it can be merged

Should be done now :+1:

PR #147, opened 2021-09-22
Closes #138

### LINKED ISSUE #138: OMTypedParser does not understand variable length arrays
I am using OMPython version 3.3.0, Python version 3.8.  When calling "getComponents", the OMTypedParser breaks if a parameter has a variable length array.  See below

**test.py**
```python


if __name__ == "__main__":
    model = """
    model Test
      parameter Real TEST[:,2] = [0,1;100,2];
    end Test;
    """

    omc = OMCSessionZMQ()
    omc.sendExpression("loadString(\"{}\")".format(model))

    # Breaks
    print(omc.sendExpression("getComponents(Test)"))
```

**Error:**
```
  File ".../test.py", line 15, in <module>
    print(omc.sendExpression("getComponents(Test)"))
  File ".../venv/lib/python3.8/site-packages/OMPython/__init__.py", line 778, in sendExpression
    answer = OMTypedParser.parseString(result)
  File ".../venv/lib/python3.8/site-packages/OMPython/OMTypedParser.py", line 120, in parseString
    res = omcGrammar.parseString(string)
  File ".../venv/lib/python3.8/site-packages/pyparsing.py", line 1955, in parseString
    raise exc
  File ".../venv/lib/python3.8/site-packages/pyparsing.py", line 3814, in parseImpl
    raise ParseException(instring, loc, self.errmsg, self)
pyparsing.ParseException: Expected end of text, found '{'  (at char 0), (line:1, col:1)
```

If I replace ":" with "2", the issue goes away and returns the expected result.

As a workaround, I am manually calling the parser and replacing ":" with "__UNBOUNDED__" before calling "OMTypedParser.parseString".

ericchapman:
Added support for variable length array ":"

### DIFF
--- a/OMPython/OMTypedParser.py
+++ b/OMPython/OMTypedParser.py
@@ -94,6 +94,7 @@ TRUE = Keyword("true").setParseAction(replaceWith(True))
 FALSE = Keyword("false").setParseAction(replaceWith(False))
 NONE = (Keyword("NONE") + Suppress("(") + Suppress(")")).setParseAction(replaceWith(None))
 SOME = (Suppress(Keyword("SOME")) + Suppress("(") + omcValue + Suppress(")"))
+VARIABLE_ARRAY = Keyword(":").setParseAction(replaceWith(":"))
 
 omcString = QuotedString(quoteChar='"', escChar='\\', multiline=True).setParseAction(convertString)
 omcNumber = Combine(Optional('-') + ('0' | Word('123456789', nums)) +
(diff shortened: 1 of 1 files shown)


sjoelund (changes requested):
This is not the correct fix. Modelica values do not contain `:` as expressions. And from the linked issue, the problem is that getComponents() is not part of the typed API. Perhaps it could be parsed if it returned `":"` instead of `:`, but that might require changes to OMEdit as well (and would be an issue for https://github.com/OpenModelica/OpenModelica)

ericchapman:
@sjoelund Please note this is the issue I was trying to address https://github.com/OpenModelica/OMPython/issues/138
 
I would argue this is a completely valid thing to do 
```python


if __name__ == "__main__":
    model = """
    model Test
      parameter Real TEST[:,2] = [0,1;100,2];
    end Test;
    """

    omc = OMCSessionZMQ()
    omc.sendExpression("loadString(\"{}\")".format(model))

    # Breaks
    print(omc.sendExpression("getComponents(Test)"))
```

ericchapman:
@sjoelund Oh sorry.  I see you referenced my 'getComponents("Test")' call so you did see it.  So are you saying that OMPython will not support making this call as it is not a part of the "typed" API?  I had a very complicated workaround that required me to set "parsed=False" on the "sendExpression" call and then replace all ":" with a string so the parser would work, and then I had to replace it back after the parse.

ericchapman:
@sjoelund As far as I could tell, OMPython doesn't handle ":" at all.  This was just a small example to replicate.  Is OMEdit even able to handle it?

sjoelund:
The problem is that the untyped API has many such special cases that are illegal code. We have many such instances and also cases where AST fragments are returned.

This particular case looks like something that is not all too hard to change in the compiler itself. Perhaps even to typed API and in that case OMEdit could get faster performance by skipping a lot parsing and type checking.

I would guess that OMPython.OMParser could parse this as it was implemented to parse anything in the untyped API. It's quite annoying to use though, especially for things that are returned in a structured way.

adeas31:
@ericchapman this needs to rebased to the new code so we can test it.

adeas31:
Moved this to milestone 5.0.0. If there is no update in future then close this one.

PR #155, opened 2022-04-13
add setCurrDir function

Arinomo:
add setCurrDir function to update CurrDir attribute and working path of getconn


OMPython/__init__.py
```diff
@@ -1773,6 +1773,19 @@ def getLinearMatrixValues(self,matrix):
             return tmpmatrix
         else:
             return np.zeros((0,0))
+        
+    def setCurrDir(mo, newPath):
```
adeas31, line 1777: Can you change this to `def setCurrDir(self, newPath):` so its consistent with the rest of the code.

OMPython/__init__.py
```diff
@@ -1773,6 +1773,19 @@ def getLinearMatrixValues(self,matrix):
...
+        path = newPath.replace('\\', '/')
+
+        if path[-1] == '/':
+            path = path[:-1]
+
+        if not os.path.exists(path):
+            raise Exception("Error: path does not exist")
+
+        mo.currDir = path
```
adeas31, line 1786: Set `currDir` if `cd` is successful.

OMPython/__init__.py
```diff
@@ -1773,6 +1773,19 @@ def getLinearMatrixValues(self,matrix):
...
+        mo.requestApi('cd', entity=path)
```
adeas31, line 1787: Check if the returned string of `cd` is same as `path` if yes then the call is successful. There is no need to check for error via `getErrorString`

OMPython/__init__.py
```diff
@@ -1773,6 +1773,19 @@ def getLinearMatrixValues(self,matrix):
...
+        return print(mo.getconn.sendExpression('cd()'))
```
adeas31, line 1788: Why are you trying to call `cd` twice and why do you want to print it? Perhaps your intention is to call `getErroString` here. Also use `requestApi` instead of directly calling `sendExpression`.

adeas31 (changes requested):
You also need to rebase your code.

Arinomo:
Hallo @adeas31 thanks for the review, but i guess the PR is not relevant anymore.

I thought one need to change the ```currDir``` in order to simulate the ```ModelicaSystem```when the working directory is change in python (see [#156](https://github.com/OpenModelica/OMPython/issues/156))

Issue #151, opened 2022-04-07
ModelicaSystem.simulate() does not handle 'outputFormat' and 'variableFilter' options correctly.


Nobby-n:
Overview
----------
The argument '-override' of (model).exe has the following syntax

(model).exe -override=startTime=0,stopTime=5,stepSize=0.01,tolerance=1e-06,solver=dassl,mdl.param1=0.5,mdl.param2=2.0,outputFormat=csv,variableFilter=. *

ModelicaSystems.simulate() will give the '-override' option if parameters or simulation options are set using setParameters() or setSimulationOptions().

In this case, 'outputFormat' and 'variableFilter' cannot be specified.
To specify 'outputFormat' and 'variableFilter', I added options to the 'simflags' argument as below, but after using setParameters() or/and setSimulationOptions(), '-override' is double argument and simulate() does not work properly.

mod.simulate(simflags='-override=startTime=0,stopTime=5,stepSize=0.01,tolerance=1e-06,solver=dassl,outputFormat=csv,variableFilter=.*')

Operating Environment
---------------------
Windows 10 Enterprise ver 20H2
OpenModelica 1.16.2 MSL3.2.3
Python 3.9.6
OMPython 3.3.0

Tentative measures
------------------
Do not use setParameters(), setSimulationOptions(), and specify all 'override' options using simFlags.

simulate() improvement suggestions
----------------------------------
Add arguments to specify 'outputFormat' and 'variableFilter' and add code to append them to the '-override' statement.

The tested 'simulate()' modifications are as follows.
```
--- C:/venv39/jupyter-lab/Lib/site-packages/OMPython/__init__.py.org	Thu Mar 17 15:14:01 2022
+++ C:/venv39/jupyter-lab/Lib/site-packages/OMPython/__init__patch.py	Fri Apr  1 18:02:55 2022
@@ -1197,13 +1197,30 @@
             return ([self.optimizeOptions.get(x,"NotExist") for x in names])

     # to simulate or re-simulate model
-    def simulate(self,resultfile=None,simflags=None):  # 11
+    ##  override option add
+    # -def simulate(self,resultfile=None,simflags=None):  # 11
+    def simulate(self,resultfile=None,simflags=None,overrideaux=None):  # 11
         """
         This method simulates model according to the simulation options.
+
+        Parameters
+        ----------
+        resultfile : str or None
+            Output file name
+
+        simflags : str or None
+            Other simulation options not '-override' parameters
+
+        overrideaux : str or None
[...]

PR #165, opened 2022-09-12
Enable the usage of a pre-built model executable when instantiating ModelicaSystem

joewa:
### Related Issues
This pull request was inspired from the discussion in issue #151 

### Purpose
- Enable the usage of a pre-built model executable when instantiating `ModelicaSystem`. This will gain some efficiency e.g. when only the parameters or the input data changes between different simulation runs.
- Apply the customization (variableFilter or overrides in general) when calling the pre-built model executable (as an alternative to the variableFilter that can be passed to ModelicaSystem or buildModel that was introduced in issue #151).

### Approach
ModelicaSystem will not to call buildModel when the `xmlFileName` of the pre-built model  is provided.
```python
mod = ModelicaSystem(
      fileName=`BouncingBall.mo`, modelName='BouncingBall',
      xmlFileName='BouncingBall_init.xml'
  )
```
The simulation can be executed with a customized set of output variables and stopTime like this:
```python
res_vars = ['h', 'v']
log_str = str(mod.simulate(
    resultfile='/tmp/bouncingball_res.mat',
    simflags=None,
    overrideaux='stopTime={0},'.format(stopTime) + 'variableFilter="'+'|'.join(res_vars)+'"'
))
```
The log_str contains the models output from stdout and stderr.

Thanks to @Nobby-n and @arun3688 for contribution and inspiration.


joewa:
I am happy to provide this PR @arun3688

adeas31:
@joewa please rebase this to the latest code.

joewa:
Please let me know when you think that the time is right @adeas31 @syntron 
The I will rebase.

syntron:
@joewa please check PR #275 - I will prepare the code in it as single PRs; however, you could check out what to expect:

Class ModelicaSystemCmd would handle the execution of commands / calling the model executable. That way, the options (simflags / new: simargs) can be handled dynamically. Thus, any option could be overwritten (override!) as needed. See especially commit 2bd6754a5b2c89894df8b17780ee3f269b3a8540 which was created for this usecase.

syntron:
@joewa feel free to work based on PR #279

joewa:
OK, will do. However it might take a week or two...

PR #178, opened 2023-10-21
Fix mistyped fileNamePrefix.

### LINKED ISSUE #176: Mistyped default fileNamePrefix
### Description
The check for the default `fileNamePrefix` is mistyped in the following line: 
https://github.com/OpenModelica/OMPython/blob/52e3dcd59fe283dcf8770297126672c22209853d/OMPython/__init__.py#L1596
which shows `"<default"` but should be `"<default>"`
This can lead to wrong paths eventually leading to problems when compiling the FMU. 
Simply adding the missing character `>` at the end of the string would fix the issue.

JavierArroyoBastida:
### Related Issues

#176 

### Purpose

Fixes #176.


casella:
Thanks for the contibution! 🎉

PR #203, opened 2024-03-07
Always call getErrorString() when verbose=True

adeas31:
If verbose=False then only call getErrorString() in case of failure.


adeas31:
supersedes #202

Issue #206, opened 2024-04-19
Simulation with Resultfile or Flags Fail to execute


NilsDressler:
### Description
When calling
`OMPython.simulaute(resultfile='/home/user/modelica/testfile0.mat')`
with a result file (and most likely also with other flags, the following log message is shown

> Failed to simulate simulation: [Errno 2] No such file or directory: '/tmp/tmp37u5_obx/Package.Simulation -r=/home/user/modelica/testfile0.mat'

### Steps to reproduce

Minimal Example that fails

```
from os.path import expanduser
from OMPython import ModelicaSystem

home_dir = expanduser("~")

model_path = home_dir + '/modelica/resultfilebug/'
model_file_name = 'BouncingBall.mo'
model_name = 'BouncingBall' 
mod = ModelicaSystem(model_path + model_file_name, model_name )

mod.buildModel()

result_file_path = home_dir + '/modelica/resultfilebug/resultfiles/'
result_file_name = 'resultfile_0.mat'

mod.simulate(resultfile=result_file_path+result_file_name)
```

Throws followingexception:

> Exception has occurred: FileNotFoundError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
> [Errno 2] No such file or directory: '/tmp/tmpj5k_1p6k/BouncingBall -r=~/modelica/resultfilebug/resultfiles/resultfile_0.mat'

Even though the file exists in my file system

```
ls /tmp/tmpqmbz5d57/BouncingBall
/tmp/tmpqmbz5d57/BouncingBall
```

Running 

`user@machine:~$  /tmp/tmpqmbz5d57/BouncingBall -r=/home/user/modelica/resultfilebug/resultfiles/resultfile_0.mat`

> LOG_ASSERT        | debug   | simulation_input_xml.c: Error: can not read file BouncingBall_init.xml as setup file to the generated simulation code.
> Execution failed!

While

`user@machine:/tmp/tmpq5rla9at$ ./BouncingBall -r=/home/user/modelica/resultfilebug/resultfiles/resultfile_0.mat`

> LOG_SUCCESS       | info    | The initialization finished successfully without homotopy method.
> LOG_SUCCESS       | info    | The simulation finished successfully.

### Expected behavior

The model should just build and return the following log messages

```
LOG_SUCCESS       | info    | The initialization finished successfully without homotopy method.
LOG_SUCCESS       | info    | The simulation finished successfully.
```

### Version and OS
 - Python Version: Python 3.10.12
[...]

PR #214, opened 2024-05-07
Changed to use list instead of string for subprocess command.

j-emils:
### Related Issues

[](https://github.com/OpenModelica/OMPython/issues/206)

### Purpose

Cannot execute the simulate method on linux.

### Approach

Adapting subprocess input.

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -1280,6 +1280,7 @@ class ModelicaSystem(object):
         currentDir = os.getcwd()
         if (os.path.exists(getExeFile)):
             cmd = getExeFile + override + csvinput + r + simflags
+            cmd = cmd.split(" ")
             #print(cmd)
             os.chdir(self.tempdir)
             if (platform.system() == "Windows"):


arun3688:
@j-emils Thanks for the fix

arun3688:
@j-emils Thanks for the fix

lafrech:
I came to report this until I realized it was fixed in master branch already. Thanks @j-emils for the fix.

A proper resolution would be to create the list in the first place rather than concatenate strings then split, but oh well.

Also, I believe the list should be split one step further to separate e.g. `-r` and the file path.

See for instance [Popen docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen):

```py
    Popen(["/usr/bin/git", "commit", "-m", "Fixes a bug."])
```

although I can't comment on the consequences of not doing it.

I did it like this as a test but really, it would be more sensible to build the list in the first place.

```py
cmd = sum([r.split("=") for r in cmd.split()], [])
```

@arun3688 considering the fixes that have been merged since last release, a new release would be greatly appreciated.

Thanks!

casella:
@j-emils do you mean you'd like to get this fix in OpenModlica 1.23.x?

arun3688:
@j-emils A new release has been made with v.3.5.2 and also the pip package has been updated with the new release `pip install OMPython==3.5.2`

j-emils:
@lafrech, I agree that a more thorough solution would be to create the list from the get go. However, the focus of this fix was to resolve the immediate issue of the code not working. In the future there is a need for refactoring the code in my opinion.

lafrech:
@j-emils I totally understand and agree
@casella done: #217

Have a nice day, guys.

PR #226, opened 2024-10-30
Om python syntron

syntron:
### Related Issues

Use OMPython with models which include external libraries (with compiled C code). Currently, the dll files are not
found if these libraries are added via the lmodel option.

### Purpose

Allow building such models.

### Approach

Adding the path based on the settings in lmodel to the environment variable.

Additionally, some cleanup of the code to execute a (Modelica) command is included (=> _run_cmd()).

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -1303,15 +1303,15 @@ class ModelicaSystem(object):
                                         dllPath = pkgpath_wver + os.pathsep + dllPath
                 my_env = os.environ.copy()
                 my_env["PATH"] = dllPath + os.pathsep + my_env["PATH"]
-                if not verbose:
-                    p = subprocess.Popen(cmd, env=my_env, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
-                else:
-                    p = subprocess.Popen(cmd, env=my_env)
             else:
-                if not verbose:
-                    p = subprocess.Popen(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
-                else:
-                    p = subprocess.Popen(cmd)
+                my_env = None
+
+            p = subprocess.Popen(cmd, env=my_env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            stdout, stderr = p.communicate()
+            if stderr:
+                logger.warning(f"OM error: {stderr.decode('ascii')}")
+            if verbose and stdout:
+                logger.info(f"OM output:\n{stdout.decode('ascii').strip()}")
             p.wait()
             p.terminate()
             os.chdir(currentDir)
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -928,6 +928,55 @@ class ModelicaSystem(object):
     def getWorkDirectory(self):
         return self.tempdir
 
+    def _run_cmd(self, cmd: list, verbose: bool = True):
+        logger.debug("Run OM command {} in {}".format(cmd, self.tempdir))
+
+        if platform.system() == "Windows":
+            omhome = os.path.join(os.environ.get("OPENMODELICAHOME"))
+            dllPath = (os.path.join(omhome, "bin")
+                       + os.pathsep + os.path.join(omhome, "lib/omc")
+                       + os.pathsep + os.path.join(omhome, "lib/omc/cpp")
+                       + os.pathsep + os.path.join(omhome, "lib/omc/omsicpp"))
+
+            # include path to resources of defined external libraries
+            for element in self.lmodel:
+                if element is not None:
+                    if isinstance(element, str):
+                        if element.endswith("package.mo"):
+                            pkgpath = element[:-10] + '/Resources/Library/'
+                            for wver in ['win32', 'win64']:
+                                pkgpath_wver = pkgpath + '/' + wver
+                                if os.path.exists(pkgpath_wver):
+                                    dllPath = pkgpath_wver + os.pathsep + dllPath
+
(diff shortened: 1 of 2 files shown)


adeas31:
Please provide the example model that doesn't work.

The compiler generates a <model_name>.bat file on Windows that contains list of all the dependencies. We should read from that file and set the process environment accordingly.

syntron:
> Please provide the example model that doesn't work.
> 
> The compiler generates a <model_name>.bat file on Windows that contains list of all the dependencies. We should read from that file and set the process environment accordingly.

Thanks for reviewing the code!

Thus, your proposal would be to read the *.bat file and set the path based on the content in this file?

My change is based on the interpretation, that the path is build using the variable dllPath which is created within Python; thus, I could add the needed content directly at this point.

Some background:
I try to use SDF tables within OpenModelica (https://github.com/ScientificDataFormat/SDF-Modelica). This works fine if I use the tool directly; however using OMPython I get no direct error - it just fails because the executable does not find the needed library (ModelicaSDF.dll).

The attached files are:
(1) python test script - [test_OMPython.py.txt](https://github.com/user-attachments/files/17617229/test_OMPython.py.txt)
(2) output using master (SHA1 = c6a204e5d6f5fe83dd3b746a403480ae2c82d09a) - [syntron_OMPython-A.txt](https://github.com/user-attachments/files/17617232/syntron_OMPython-A.txt)
(3) output using the new commits - [syntron_OMPython-B.txt](https://github.com/user-attachments/files/17617233/syntron_OMPython-B.txt)

Besides the above listed point to run the model with the external library in OMPython, I also would like to create an FMU. However, this also fails and I expect a similar root cause ...

adeas31:
> > Please provide the example model that doesn't work.
> > The compiler generates a <model_name>.bat file on Windows that contains list of all the dependencies. We should read from that file and set the process environment accordingly.
> 
> Thanks for reviewing the code!
> 
> Thus, your proposal would be to read the *.bat file and set the path based on the content in this file?

Yes.

> 
> My change is based on the interpretation, that the path is build using the variable dllPath which is created within Python; thus, I could add the needed content directly at this point.
> 
> Some background: I try to use SDF tables within OpenModelica (https://github.com/ScientificDataFormat/SDF-Modelica). This works fine if I use the tool directly; however using OMPython I get no direct error - it just fails because the executable does not find the needed library (ModelicaSDF.dll).

I suppose you mean it works fine from OMEdit, right?
OMEdit reads the batch file and sets the process path correctly. We need to do the same for OMPython.

> 
> The attached files are: (1) python test script - [test_OMPython.py.txt](https://github.com/user-attachments/files/17617229/test_OMPython.py.txt) (2) output using master (SHA1 = [c6a204e](https://github.com/OpenModelica/OMPython/commit/c6a204e5d6f5fe83dd3b746a403480ae2c82d09a)) - [syntron_OMPython-A.txt](https://github.com/user-attachments/files/17617232/syntron_OMPython-A.txt) (3) output using the new commits - [syntron_OMPython-B.txt](https://github.com/user-attachments/files/17617233/syntron_OMPython-B.txt)
> 
> Besides the above listed point to run the model with the external library in OMPython, I also would like to create an FMU. However, this also fails and I expect a similar root cause ...

arun3688:
@adeas31 OMPython does use the  .exe file to simulate the model, may be we should just change the API to use the `.bat` to simulate.

@syntron Can you provide your python script on how you call the API, that could help to solve the issue

adeas31:
> @adeas31 OMPython does use the .exe file to simulate the model, may be we should just change the API to use the `.bat` to simulate.

OMPython ModelicaSystem calls `buildModel` and then runs the generated executable. `buildModel` will create the batch file which should be used to set the environment for executable.

arun3688:
> OMPython ModelicaSystem calls `buildModel` and then runs the generated executable. `buildModel` will create the batch file which should be used to set the environment for executable.

sure i will change the simulate API to use the batch file to set up the environment

syntron:
> @syntron Can you provide your python script on how you call the API, that could help to solve the issue

The example script is attached to my comment above

syntron:
> > OMPython ModelicaSystem calls `buildModel` and then runs the generated executable. `buildModel` will create the batch file which should be used to set the environment for executable.
> 
> sure i will change the simulate API to use the batch file to set up the environment

OK; I'm ready to test the change.

Would the definition of _run_cmd() still be interesting? It would run for *.exe and *.bat files ...

syntron:
From [[arun3688](https://github.com/arun3688)](https://github.com/OpenModelica/OMPython/pull/226#issuecomment-2454493822):
> > OMPython ModelicaSystem calls `buildModel` and then runs the generated executable. `buildModel` will create the batch file which should be used to set the environment for executable.
> 
> sure i will change the simulate API to use the batch file to set up the environment

From https://github.com/OpenModelica/OMPython/pull/226#issuecomment-2454435178:
> Besides the above listed point to run the model with the external library in OMPython, I also would like to create an FMU. However, this also fails and I expect a similar root cause ...

Would something like this also help for FMU creation?

arun3688:
@adeas31 Yes that is true but this is basically some refactoring of code to run the executable for `simulation and linearization`, i will do the batch file implementation in a separate PR

syntron:
> @syntron I have added some tests and the PR looks good thanks for the contribution

Thanks for the merge! Any hint where to look to get FMU compilation working for external libraries?

syntron:
> Any hint where to look to get FMU compilation working for external libraries?

I just tested it - with the development version 1.25 (exactly: OpenModelica-v1.25.0-dev-106-g982563026f-64bit.exe) it is working fine!

PR #238, opened 2024-12-11
Omhome as argument

syntron:
allow to set omhome as argument to ModelicaSystem

* no need to modify the environment on the system
* changes mainly in OMCSessionHelper and going up to ModelicaSystem
* includes some cleanup (muteable arguments)


adeas31 (approved):
LGTM

PR #239, opened 2025-03-06
Add session kwarg to ModelicaSystem constructor

AxelMatstoms:
If the session kwarg is specified, the ModelicaSystem will use that session instead of creating a new one. This is useful for creating a ModelicaSystem for a model that is already loaded in an already existing session.

### Purpose
The purpose of the change is to allow usage of the high level ModelicaSystem API around models that are already loaded, or created through `createModel` etc in an existing session.

### Approach
By adding a session keyword argument to the ModelicaSystem constructor, users can use the ModelicaSystem constructor as follows to construct a ModelicaSystem around an already loaded model from an existing session:
```python
omc = OMCSessionZMQ()
omc.sendExpression("createModel(ExampleModel)")
...
model = ModelicaSystem(None, "ExampleModel", session=omc)
model.simulate()
```

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -821,7 +821,7 @@ class ModelicaSystem(object):
     def __init__(self, fileName=None, modelName=None, lmodel=None,
                  useCorba=False, commandLineOptions=None,
                  variableFilter=None, customBuildDirectory=None, verbose=True, raiseerrors=False,
-                 omhome: str = None):  # 1
+                 omhome: str = None, session: OMCSessionBase = None):  # 1
         """
         "constructor"
         It initializes to load file and build a model, generating object, exe, xml, mat, and json files. etc. It can be called :
@@ -831,11 +831,14 @@ class ModelicaSystem(object):
         Note: If the model file is not in the current working directory, then the path where file is located must be included together with file name. Besides, if the Modelica model contains several different models within the same package, then in order to build the specific model, in second argument, user must put the package name with dot(.) followed by specific model name.
         ex: myModel = ModelicaSystem("ModelicaModel.mo", "modelName")
         """
+        if session is not None:
+            self.getconn = session
+        elif useCorba:
+            self.getconn = OMCSession(omhome=omhome)
+        else:
+            self.getconn = OMCSessionZMQ(omhome=omhome)
+
         if fileName is None and modelName is None and not lmodel:  # all None
-            if useCorba:
-                self.getconn = OMCSession(omhome=omhome)
-            else:
-                self.getconn = OMCSessionZMQ(omhome=omhome)
             return
 
         self.tree = None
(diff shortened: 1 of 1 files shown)


AxelMatstoms:
@adeas31 Are there any other concerns with the ownership other than the `__del__` methods that would prevent sharing ownership of a single session? Otherwise, an easy solution would be to remove the `__del__` method from `ModelicaSystem`. The `__del__` method on the session will be called when the `ModelicaSystem` goes out of scope anyways.

If not I could update the PR removing the del method. I also realized I forgot to set the session in the case the init method was called with zero positional arguments.

AxelMatstoms:
Updated to remove the deleter from `ModelicaSystem`. I've done some testing for this. Previously the session deleter would run multiple times for the same process pid. Removing the deleter from `ModelicaSystem` ensures it is run at most once (since Python doesn't guarantee that the del method will be called on objects that still exist at the end of the program[^1]). 

[^1]: [https://docs.python.org/3/reference/datamodel.html#object.\_\_del\_\_](https://docs.python.org/3/reference/datamodel.html#object.__del__)

adeas31:
Maybe we should not remove the `__del__` method, it is fine to call it once ModelicaSystem goes out of scope.

I tested the following and it works fine,

```python
omc = OMCSessionZMQ()
omc.sendExpression("createModel(ExampleModel)")
model = ModelicaSystem(None, "ExampleModel", session=omc)
model.buildModel()
```

However, the following doesn't work,

```python
omc = OMCSessionZMQ()
omc.sendExpression("createModel(ExampleModel)")
model = ModelicaSystem(None, "ExampleModel", session=omc)
omc.sendExpression("quit()")
model.buildModel()
```

This gives the exception error,

```
OMPython - ERROR - Exception <class 'Exception'> raised: Process Exited, No connection with OMC. Create a new instance of OMCSession
  File "<stdin>", line 1, in <module>
  File "C:\OpenModelica\OMPython\OMPython\__init__.py", line 1038, in buildModel
    self._check_error()
  File "C:\OpenModelica\OMPython\OMPython\__init__.py", line 1013, in _check_error
    errstr = self.getconn.sendExpression("getErrorString()")
  File "C:\OpenModelica\OMPython\OMPython\__init__.py", line 813, in sendExpression
    raise Exception("Process Exited, No connection with OMC. Create a new instance of OMCSession")
Exception: Process Exited, No connection with OMC. Create a new instance of OMCSession
```

I guess it is expected output as the `omc` session that `ModelicaSystem` is using has been closed. The only way forward from here is to create a new `omc` session and new `ModelicaSystem`.

@arun3688 what do you think?

AxelMatstoms:
I think the problem with having the del method on ModelicaSystem is that it gets called twice: once when the ModelicaSystem goes out of scope, and once when the session goes out of scope.

I tried the following on the upstream master branch:
```python
from OMPython import ModelicaSystem, OMCSessionBase

deleter = OMCSessionBase.__del__
def new_deleter(self):
    print("__del__:", self._omc_process.pid)
    deleter(self)
OMCSessionBase.__del__ = new_deleter

model = ModelicaSystem(None, "Modelica.Blocks.Examples.PID_Controller")
```
which gives this output showing that the deleter runs twice. Removing the deleter from ModelicaSystem fixes this.
```
__del__: 105196
__del__: 105196
```

PR #242, opened 2025-04-05
Add debug logging to ModelicaSystem sendExpression

ondras12345:
I have just spent several hours debugging why `ModelicaSystem()` cannot load my custom package.
It turns out that it does not support relative paths, because it `cd()`s to a tmp directory.
However, that is really hard to see without appropriate logging.

This PR allows the user to see exactly what's happening by enabling debug logging:
```python3
import logging
logging.basicConfig(level=logging.DEBUG)
logging.getLogger("OMPython").setLevel(logging.DEBUG)
```


lochel (approved):
Very good 👍

PR #243, opened 2025-04-06
Fix linearize() simflags must start with space

ondras12345:
Previously, the flags supplied by the user in `simflags` would get appended to automatically generated flags without a space in a call like this:
```python3
(A, B, C, D) = mod.linearize(simflags="-r=file.mat")
```
A workaround is to start simflags with a space:
```python3
(A, B, C, D) = mod.linearize(simflags=" -r=file.mat")
```

This PR removes the need for the workaround.

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -1651,6 +1651,8 @@ class ModelicaSystem(object):
 
         if simflags is None:
             simflags = ""
+        else:
+            simflags = " " + simflags
 
         if (os.path.exists(getExeFile)):
             cmd = getExeFile + linruntime + override + csvinput + simflags


ondras12345:
I have already done that when I was submitting https://github.com/OpenModelica/OMPython/pull/242. I see that the licese/cla check has passed, too. Do I need to do it for each PR separately? If so, how? Last time, the CLAassistant bot sent me a link.

lochel:
No, the cla is fine. The other jobs need to be started manually for whatever reason. I just did so.

lochel (approved):
A bit hacky, but seems to be alright.

Issue #245, opened 2025-04-07
ModelicaSystem cannot load model from relative path


ondras12345:
### Description
The documentation says that
https://github.com/OpenModelica/OMPython/blob/5904e4eec81d215bc677dc45f407ed9e2d304733/OMPython/__init__.py#L660

However, this doesn't work in practice:
```
$ cd /usr/share/doc/omc/testmodels 
$ ipython3                                   
/usr/lib/python3/dist-packages/IPython/core/interactiveshell.py:915: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn(
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from OMPython import ModelicaSystem

In [2]: mod = ModelicaSystem("BouncingBall.mo", "BouncingBall")
2025-04-07 11:31:12,386 - OMPython - ERROR - Error: Failed to load package BouncingBall (default) using MODELICAPATH /home/ondra/.openmodelica/libraries/.
Error: Class BouncingBall not found in scope <top>.

2025-04-07 11:31:12,386 - OMPython - ERROR - XML file not generated: 

In [3]: from pathlib import Path

In [4]: mod = ModelicaSystem(str(Path("BouncingBall.mo").resolve()), "BouncingBall")

In [5]: 
```

### Version and OS
 - Python Version: 3.12.3
 - OMPython Version: 5904e4e
 - OpenModelica Version: 1.25.0
 - OS: Ubuntu 24.04.2 LTS

### Additional Context
I think I'll try to fix this myself and submit a pull request.

PR #247, opened 2025-04-09
Fix ModelicaSystem cannot load from relative path

ondras12345:
This fixes https://github.com/OpenModelica/OMPython/issues/245


arun3688:
@ondras12345 Good work

PR #246, opened 2025-04-07
Improve code format & add tests

ondras12345:
This PR fixes a couple of code format issues, like inconsistent indentation (use of 2 spaces instead of 4), unused variables and imports, and PEP8 violations detected by flake8.
It also adds a couple of tests, removes Python2 compatibility code, switches from `setup.py` to `pyproject.toml` and configures Github actions CI to run `flake8` linter.

If you don't want all of these changes, feel free to let me know, I can cherry-pick some commits and make a new PR.

PR #249, opened 2025-04-12
Clean up code & improve test coverage

ondras12345:
This PR tries to clean up the code by:
- using f-strings instead of old %-style format strings and `.format()`,
- using pathlib.Path and `.as_posix()` instead of `os.path.*` and manually replacing `\` with `/`,
- dropping legacy code that was needed for Python2 compatibility,
- dropping unreachable code.

There should be no significant changes in behavior (except for the last commit "Fix createCSVData on Windows", which removes an extra newline at the end of the csv file).

It also adds tests, improving `OMPython/__init__.py` coverage to 73% on Linux.

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -516,7 +516,7 @@ class OMCSessionZMQ(OMCSessionBase):
         if sys.platform == 'win32':
             self._omc_command = omc_path_and_args_list
         else:
-            self._omc_command = ' '.join([shlex.quote(a) if (sys.version_info > (3, 0)) else a for a in omc_path_and_args_list])
+            self._omc_command = ' '.join([shlex.quote(a) for a in omc_path_and_args_list])
 
         return self._omc_command
 
@@ -551,7 +551,7 @@ class OMCSessionZMQ(OMCSessionBase):
         while True:
             if self._dockerCid:
                 try:
-                    self._port = subprocess.check_output(["docker", "exec", self._dockerCid, "cat", self._port_file], stderr=subprocess.DEVNULL if (sys.version_info > (3, 0)) else subprocess.STDOUT).decode().strip()
+                    self._port = subprocess.check_output(["docker", "exec", self._dockerCid, "cat", self._port_file], stderr=subprocess.DEVNULL).decode().strip()
                     break
                 except Exception:
                     pass
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -1464,7 +1464,7 @@ class ModelicaSystem:
             l.append(a)
 
         self.csvFile = (pathlib.Path(self.tempdir) / f'{self.modelName}.csv').as_posix()
-        with open(self.csvFile, "w") as f:
+        with open(self.csvFile, "w", newline="") as f:
             writer = csv.writer(f, delimiter='\n')
             writer.writerow(l)
         f.close()
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -537,10 +537,7 @@ class OMCSessionZMQ(OMCSessionBase):
         raise ValueError("Cannot find OpenModelica executable, please install from openmodelica.org")
 
     def _get_omc_path(self) -> pathlib.Path:
-        omc = self.omhome / "bin" / "omc"
-        if not omc.exists():
-            raise FileNotFoundError(f"The OpenModelica compiler is missing in the System path ({omc}), please install it")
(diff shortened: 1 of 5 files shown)


OMPython/__init__.py
```diff
@@ -279,7 +279,7 @@ def _set_omc_command(self, omc_path_and_args_list):
         if sys.platform == 'win32':
             self._omc_command = omc_path_and_args_list
         else:
-            self._omc_command = ' '.join([shlex.quote(a) if (sys.version_info > (3, 0)) else a for a in omc_path_and_args_list])
+            self._omc_command = ' '.join([shlex.quote(a) for a in omc_path_and_args_list])
```
syntron, line 279: would this work also without the join()? - it is just the command used for subprocess - this works well as a list; we only have to check that all items are quoted

syntron:
@ondras12345 I realy like the work you have done; especially the string formating! I was thinking about this myself but stopped as I was not sure how much of it is needed for Python 2.x compatibility. However, as this is droped, these changes make the code more readable!

OMPython/__init__.py
ondras12345: I have just tried it, and it doesn't seem to like the change:
```
/home/ondra/source/repos/OMPython/tests/test_ZMQ.py:17: in __init__
    self.om = OMPython.OMCSessionZMQ()
/home/ondra/source/repos/OMPython/OMPython/__init__.py:551: in __init__
    self._connect_to_omc(timeout)
/home/ondra/source/repos/OMPython/OMPython/__init__.py:581: in _connect_to_omc
    raise Exception(f"OMC Server did not start (timeout={timeout}). Could not open file {self._port_file}"
)
E   Exception: OMC Server did not start (timeout=10.0). Could not open file /tmp/openmodelica.ondra.port.c431836ad07c4d41b06cd1eee41cffff
```

However, I don't know why. At first glance, it should work. E.g. this works:
```
In [1]: import subprocess

In [2]: subprocess.Popen(['/usr/bin/omc', '--help'])
Out[2]: <Popen: returncode: None args: ['/usr/bin/omc', '--help']>

OpenModelica Compiler OpenModelica 1.25.0
Copyright © 2019 Open Source Modelica Consortium (OSMC)
Distributed under OMSC-PL and GPL, see www.openmodelica.org

Usage: omc [Options] (Model.mo | Script.mos) [Libraries | .mo-files] 
[...]
```

I will investigate further tomorrow.

ondras12345:
> how much of it is needed for Python 2.x compatibility. However, as this is droped

I didn't even notice the README said the library still supports Python2, and since setup.py said `python_requires='>=3.8'`, I assumed that's the oldest version we need to support and started removing python2 workarounds in https://github.com/OpenModelica/OMPython/pull/246.

Also as a side note, if support for 3.8 is declared in `pyproject.toml`, we should probably run tests @ 3.8 and not just 3.10 and 3.12:
https://github.com/OpenModelica/OMPython/blob/bfb115376fa4090592566fd8d4dfafd2dc27a238/.github/workflows/Test.yml#L15
To be fair, I didn't test my changes with Python 3.8. I should go do that tomorrow...

OMPython/__init__.py
ondras12345: Fixed in [20c24dc](https://github.com/OpenModelica/OMPython/pull/249/commits/20c24dc666be595608246f8d2e200e84e2a0c9bb)
The reason it didn't work was that Popen can't take a list with shell=True.

ondras12345:
I tried to run a set of tests with python 3.8, and it doesn't even install. And it's totally my fault, because it fails to parse pyproject.toml, which was introduced by me in https://github.com/OpenModelica/OMPython/pull/246. Error log:
https://github.com/ondras12345/OMPython/actions/runs/14511691030/job/40711663255
I'll see what I can do about it.

syntron:
> and they are passing. It should be safe to say python >= 3.10 is supported. It's not like we need the new features introduced in 3.11 and 3.12.

I support staying with Python >= 3.10 as this version is working well in the tests; why should we limit the possible user range without a really need for such a change?

ondras12345:
I have rebased my branch onto master and fixed the conflicts (there was a lot of them).
If the tests pass (especially the windows ones, I already ran linux tests locally), this PR should be ready for review @adeas31

arun3688 (approved):
looks good

adeas31 (approved):
LGTM. Thanks!

PR #250, opened 2025-04-15
Refactor omc session classes

syntron:
merge OMCSessionHelper into OMCSessionZMQ

define execute as deprecated

merge (parts of) OMCSessionBase into OMCSessionZMQ

* class OMCSessionBase only contains ask() and derived functions to call OM
* the moved functions were using variables of class OMCSessionZMQ
* this change splits the class definitions into
   (1) code to connect to OM => OMCSessionBase
   (2) code to execute OM / connect to running OM => OMCSessionZMQ

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -101,38 +101,6 @@ class DummyPopen():
         return self.process.wait(timeout=timeout)
 
 
-class OMCSessionHelper:
-    def __init__(self, omhome: str = None):
-        self.omhome = None
-
-        # use the provided path
-        if omhome is not None:
-            self.omhome = omhome
-            return
-
-        # check the environment variable
-        omhome = os.environ.get('OPENMODELICAHOME')
-        if omhome is not None:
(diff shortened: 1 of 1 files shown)


arun3688:
@syntron can you resolve the conflicts and update the PR

syntron:
> @syntron can you resolve the conflicts and update the PR

Done

syntron:
@adeas31 Maybe in another PR we can also remove `OMCSessionBase` as it doesn't serve any purpose now.

I could create such a commit; however, my proposal would be to only add some cleanup as the current structure quite nicely splits
(1) OM commands => OMCSessionBase (only needs an implementation of sendExpression()) and
(2) code needed to connect to an OM session => OMCSesssionZMQ

This way, it would be possible to keep OMCSessionZMQ clean. Furthermore, it could allow to split it into code which is needed to run OM locally and the code which is related to docker based execution. I expect, that the docker case would be something which could run on top of the local execution.

PR #257, opened 2025-04-17
Refactor file structure

syntron:
### Related Issues

See #254

### Purpose

Separate the differnet classes / parser files such that specific commits touch only the corresponding file.
Use init.py only for the package related definitions (public interface of the package)

### Approach

* move classes from init.py into separate files (=> OMCSession; ModelicaSystem)
* move OMParser/init.py into OMParser.py as it is just a directory

This is splitted into several commits such that the main change (move class definitions) can easily be rebased / redone

There are some smaller changes which I would add on top of this change / after #249

PR #258, opened 2025-04-17
Some more code cleanups

syntron:
additional cleanups which should be added before #257
* Remove unreachable code
* cleanup old code for Python 2.x
* fix OMCSessionBase - the variables self.readonly and self.omc_cache are needed here
* [ModelicaSystem] fix True in docstring
* [OMCSessionBase] fix usage of excaptions - replace ex.message() by ex.msg()


syntron:
see also discussion in #250 - small changes to make OMCSessionBase and OMCSessionZMQ independent of each other

PR #263, opened 2025-04-23
Suppress PEP warnings

syntron:
### Purpose

PyCharm (and other IDE's?) are highlighting a lot of warnings in the OMPython files (default settings)

This PR reduces the number of warnings by using corresponding '# noinspection' comments.

* unresolved references (in docstrings; due to old syntax based on non-class methods) => could be handled by the docstring cleanup planned by @ondras12345
* suppress name schema warnings - upper letters in function names

### Approach

just use '# noinspection' ...

This PR is defined on top of #257


syntron:
I'm no longer sure that this is a good idea - this hides the (valid) messages of the linter; I think if these should be suppressed, it is something each developer could do on his/her own. Especially if done on the level of a class ...

PR #265, opened 2025-04-25
Refactor createCSVData

ondras12345:
I found the old code really hard to understand, so I decided to rewrite it from scratch. It should behave exactly the same (and the tests in `test_ModelicaSystem` prove that).


arun3688 (approved):
looks good

PR #266, opened 2025-04-25
Make `ModelicaSystem.linearize()` return all values

ondras12345:
I think it is a shame to get the `x0` and `u0` values from OpenModelica and throw them away without returning them to the user. Thus, I have changed `ModelicaSystem.linearize()` to return a dataclass with all the available information.

To avoid breaking code that relies on the old return value, I have made it possible to unpack the dataclass into `(A,B,C,D)`, as well as to use indexes 0-3 to access the matrices.
However, there are still cases where the new return value behaves differently, e.g. if someone passed the old result straight to `json.dumps()`.

An alternative approach would be to preserve the old return value and add a new function:
```python3
def linearize_new([...]) -> LinearizationResult:
    ...

def linearize(*args, **kwargs):
    r = linearize_new(*args, **kwargs)
    return list(r)
```

It is up to you to decide whether changing the return value like this is acceptable. If you decide to go with the alternative approach outlined above, I'll rework this pull request.

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -43,6 +43,8 @@ import xml.etree.ElementTree as ET
 import numpy as np
 import importlib
 import pathlib
+from dataclasses import dataclass
+from typing import Optional
 
 from OMPython.OMCSession import OMCSessionBase, OMCSessionZMQ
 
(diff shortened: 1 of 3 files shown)


arun3688 (approved):
The changes looks good

Issue #260, opened 2025-04-19
Inconsistent getter behavior when requested value does not exist


ondras12345:
Example:
```
In [1]: import OMPython

In [2]: mod = OMPython.ModelicaSystem("/usr/share/doc/omc/testmodels/BouncingBall.mo", "BouncingBall", raiseerrors=True)

In [3]: mod.getOutputs("ThisOutputDoesNotExist")
Out[3]: ['NotExist']

In [4]: mod.simulate()

In [5]: mod.getOutputs("ThisOutputDoesNotExist")
Out[5]: ('ThisOutputDoesNotExist', ' is not Output')
```

In my opinion, it would be best to raise an exception (maybe KeyError?) in cases like this instead of returning invalid values. However, that would be an incompatible API change.

I'm currently trying to improve the docstrings for methods in ModelicaSystem. Should I document these weird quirks?

syntron:
> I'm currently trying to improve the docstrings for methods in ModelicaSystem. Should I document these weird quirks?

@ondras12345 I'm all for an exception - possible use raise ModelicaSystemException? I think it is important to tell the user as soon as possible if some function cannot work / will result in invalid data => fail early, fail hard ... - see also [https://stackoverflow.com/questions/2807241/what-does-the-expression-fail-early-mean-and-when-would-you-want-to-do-so](url)

PR #267, opened 2025-04-25
Improve docstrings

ondras12345:
This PR tries to improve the docstrings of some of ModelicaSystem's methods.
The new docstrings follow the [Google Style Docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) (I hope).

Currently I don't have time to do this for all methods.
Also, I have encountered some questions that should be answered before docstrings for the other methods are written: https://github.com/OpenModelica/OMPython/issues/260.

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -498,19 +498,6 @@ class ModelicaSystem:
         elif isinstance(names, list):
             return ([self.paramlist.get(x, "NotExist") for x in names])
 
-    def getlinearParameters(self, names=None):  # 5
-        """
-        This method returns dict. The key is parameter names and value is corresponding parameter value.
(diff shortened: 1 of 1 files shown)


adeas31 (approved):
> Currently I don't have time to do this for all methods.
> Also, I have encountered some questions that should be answered before docstrings for the other methods are written: https://github.com/OpenModelica/OMPython/issues/260.

Lets just push this in for now. We can improve the docstrings of other methods later on.

ondras12345:
Google-style docstrings, as stated in the description of this PR. However, I don't feel strongly about it. You can change it to numpy-style if you think it's better.

syntron:
@ondras12345 sorry, this was so clear at the top that I missed it; I will have to check this style as I'm not familiar with it

PR #269, opened 2025-04-25
Remove logging

syntron:
### Related Issues

logging should be defined in the calling script

reason: the calling script should define how logging is done; if these settings are added here and the caller adds another log handle (with possibly different settings), all log information could be printed twice or not at all

### Approach

* cleanup log definnitions in __ init __.py (first commit)
* the second commit (re)adds the log definition as a method which could be called if needed

### DIFF
--- a/OMPython/__init__.py
+++ b/OMPython/__init__.py
@@ -36,26 +36,8 @@ __license__ = """
  CONDITIONS OF OSMC-PL.
 """
 
-import logging
-
 from OMPython.OMCSession import OMCSessionBase, OMCSessionZMQ
 from OMPython.ModelicaSystem import ModelicaSystem, ModelicaSystemError, LinearizationResult
-
-# Logger Defined
-logger = logging.getLogger('OMPython')
-logger.setLevel(logging.DEBUG)
-# create console handler with a higher log level
-logger_console_handler = logging.StreamHandler()
-logger_console_handler.setLevel(logging.INFO)
-
-# create formatter and add it to the handlers
-logger_formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
-logger_console_handler.setFormatter(logger_formatter)
-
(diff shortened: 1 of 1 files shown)


adeas31:
I think we should define logger and perhaps should do it like this `logger = logging.getLogger(__name__)`. Python logger is a singleton so there shouldn't be an issue of duplicate messages.

We should not configure logger i.e., remove lines like `logger.setLevel(logging.DEBUG)` and `logger.setLevel(logging.WARNING)` (btw I don't know why we are first setting the level to debug and then to warning, weird code).

In principle, we should only define and then log messages. Then its up to the user to configure and consume log messages.

I am not sure how we can fix `StreamHandler`, this is perhaps needed to log messages that comes from omc stream output in some cases.

ondras12345:
> I think we should define logger and perhaps should do it like this `logger = logging.getLogger(__name__)`

I believe the individual modules already do this. E.g. ModelicaSystem:
https://github.com/OpenModelica/OMPython/blob/98174c38364c68c76944a27e05a4f8a6e266703f/OMPython/ModelicaSystem.py#L52

As I understand it, this PR only removes the configuration - that should indeed be left to the user.

adeas31:
> > I think we should define logger and perhaps should do it like this `logger = logging.getLogger(__name__)`
> 
> I believe the individual modules already do this. E.g. ModelicaSystem:
> 
> https://github.com/OpenModelica/OMPython/blob/98174c38364c68c76944a27e05a4f8a6e266703f/OMPython/ModelicaSystem.py#L52
> 
> As I understand it, this PR only removes the configuration - that should indeed be left to the user.

Yes. And we should do the same in `__init__.py`. Also remove the multiple `setLevel` calls.

And as I said earlier, not sure how to manage stream handler.

syntron:
updated to latest HEAD and removal of the second part - now it is a clean __ init __.py

PR #270, opened 2025-04-25
Parser

syntron:
There are two parsers for OM return values which are used in OMCSessionBase. This PR tries to clean up their usage - use only a single entry point for each - and add some additional cleanup:

* define OMCSessionBase._ask_with_fallback() - the only use of OMTypedParser
* fix handling of exceptions raised by pyparsing (use .msg instead of .message)


adeas31:
@syntron I removed `_ask_with_fallback` and moved the fallback code to `sendExpression`.

I think we should also remove all the APIs and `def ask`. We should just have `def sendExpression` which users can use to call any API. It is difficult to keep track of the changes done in the API.

syntron:
> I think we should also remove all the APIs and `def ask`. We should just have `def sendExpression` which users can use to call any API. It is difficult to keep track of the changes done in the API. I will make an issue about it.

@adeas31 I just added all the commits I prepared for OMCSession.py (see PR #275). It further simplifies OMCSessionBase such that is an independend class which can be used to query OMC based on a session (= OMCSessionZMQ). As side effect, this class could later be removed - however, at the moment it is as simple as possible and - thus - perhaps still useful.

Some small additional cleanups:
* use pathlib if possible / cleanup
* add exception handling and use specific exception if possible

syntron:
just noted: this now includes PR #272 ... if needed I can split this put again ...

adeas31:
> just noted: this now includes PR #272 ... if needed I can split this put again ...

Lets keep the PRs simple. Adding small chunks of code is better than adding huge block of code. Please split and make several PRs if possible.

syntron:
> > just noted: this now includes PR #272 ... if needed I can split this put again ...
> 
> Lets keep the PRs simple. Adding small chunks of code is better than adding huge block of code. Please split and make several PRs if possible.

I restored the state of this PR before my change - the modified parser branch is kept in my fork as parserOLD and as PR #280, #281 and #282

PR #273, opened 2025-04-25
Use subprocess.run

syntron:
Simplify subprocess.Popen() by using subprocess.run() in ModelicaSystem._run_cmd()

Reason: the added functionality of Popen() is not used at all in this case ...

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -1062,6 +1062,7 @@ class ModelicaSystem:
             lintime: Override linearOptions["stopTime"] value.
             simflags: A string of extra command line flags for the model
               binary.
+            timeout: Possible timeout for the execution of OM.
 
         Returns:
             A LinearizationResult object is returned. This allows several
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -310,19 +310,13 @@ class ModelicaSystem:
             my_env = None
 
         try:
-            p = subprocess.Popen(cmd, env=my_env, stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


arun3688:
@syntron please resolve the conflicts and update your PR

OMPython/ModelicaSystem.py
```diff
@@ -310,19 +310,18 @@ def _run_cmd(self, cmd: list):
...
-            stdout, stderr = p.communicate()
-
-            stdout = stdout.decode('ascii').strip()
-            stderr = stderr.decode('ascii').strip()
+            cmdres = subprocess.run(cmd, capture_output=True, text=True, env=my_env, cwd=self.tempdir,
+                                    timeout=timeout)
+            stdout = cmdres.stdout.strip()
+            stderr = cmdres.stderr.strip()
+            if cmdres.returncode != 0:
+                raise ModelicaSystemError(f"Error running command {cmd}: nonzero return code")
```
adeas31, line 318: Return the error code. It is useful sometimes.

adeas31 (changes requested):

OMPython/ModelicaSystem.py
syntron: Done in the latest version (added a1b6aa8fb63b689da0e2fdf4254eba593a19d3cb) However, please also consider that the model executable sometimes fails due to not reaching the final simulation time; while this is correctly a fail, the simulation data up to this point could be helpful! - I considered this case in #279 (see especially 341c167f85ee08f801877ad8ac671c12e12de92a).

adeas31 (approved):

PR #275, opened 2025-04-26
Test

syntron:
This is just a draft of ideas to further work on ...

... will be splitted into separate PRs if ready ...


ondras12345:
@syntron Just a note: it is easy to run tests locally:
```
# git clone ....
cd OMPython
python3 -m venv venv
. ./venv/bin/activate
pip3 install --editable .
pip3 install pytest
pytest
```
When you push a commit to this pull request, it has to create a new virtual machine, install all dependencies in it, and then run the tests and delete the VM. And it does this for two versions of python and for both linux and Windows. If you run tests locally, you won't have to wait as long for them to finish.

syntron:
@ondras12345 thanks for the hint; the problem is that I'm limited due to the distribution I use: no package for opensuse available ... thus, I have to run testing on a different computer (or on github)

syntron:
this draft PR is  (kind of) ready ... I will split it in smaller parts; however, as there are dependencies between the elements some PR will depend on each other ...

syntron:
this draft PR is  (kind of) ready ... I will split it in smaller parts; however, as there are dependencies between the elements some PR will depend on each other ...

PR #277, opened 2025-04-30
Update modelica system

syntron:
Update ModelicaSystem on top of PR #273

Changes:

* remove verbose - use logger.debug()
* remove raiseerror - use raise ModelicaSystemException()
* use pathlib.Path() if possible
* enforce lmodel is a list
* use exceptions if possible

PR #278, opened 2025-05-01
Finish error handling

syntron:
Finish update of error handling in OMSession.py and ModelicaSystem.py

Based on PR #277, #273 and #270

Changes
* Always check for errors in sendExpression()
* align logging - log data only once
* cleanup

### DIFF
--- a/OMPython/OMCSession.py
+++ b/OMPython/OMCSession.py
@@ -549,6 +549,17 @@ class OMCSessionZMQ:
             return None
         else:
             result = self._omc.recv_string()
+
+            # allways check for error
+            self._omc.send_string("getErrorString()", flags=zmq.NOBLOCK)
+            error_raw = self._omc.recv_string()
+            error_str = om_parser_typed(error_raw)
+            if error_str:
+                if "Error" in error_str:
+                    raise OMCSessionException(f"OM Error for 'sendExpression({command}, {parsed})': {error_str}")
+                else:
+                    logger.warning(f"[OM]: {error_str}")
+
             if parsed is True:
                 try:
                     return om_parser_typed(result)
--- a/OMPython/OMCSession.py
+++ b/OMPython/OMCSession.py
@@ -41,6 +41,7 @@ import json
 import os
 import pathlib
 import psutil
+import re
 import signal
 import subprocess
 import sys
@@ -325,6 +326,9 @@ class OMCSessionZMQ:
         # connect to the running omc instance using ZMQ
         self._connect_to_omc(timeout)
 
+        self._re_log_entries = None
+        self._re_log_raw = None
+
     def __del__(self):
         try:
             self.sendExpression("quit()")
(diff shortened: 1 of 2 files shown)


adeas31:
Rebase this one and see if you can use `getMessagesStringInternal()`. We should also check if expression is `getErrorString()` or `getMessagesStringInternal()` then we should not call it twice.

syntron:
@adeas31 would it be possible to merge #281 and #282 before that? Both would merge in this order without problems even if not rebased.

This would simplify the merge as this PR is based on these two. Furthermore, I will be limited in time till mid of next week ...

syntron:
@adeas31 do you have a simple way / an example to force an error? I would like to check how to evaluate the output of `getMessagesStringInternal()`

adeas31:
Run the following script with omc,

```Modelica
loadString("model M Real foo = \"str\"; annotation (uses(Modelica(version=\"4.0.0\"))); end M;");
getMessagesStringInternal();
checkModel(M);
getMessagesStringInternal();
```

equivalent OMPython code is,

```python
omc = OMCSessionZMQ()
omc.sendExpression("model M Real foo = \"str\"; annotation (uses(Modelica(version=\"4.0.0\"))); end M;")
omc.sendExpression("getMessagesStringInternal()")
omc.sendExpression("checkModel(M)")
omc.sendExpression("getMessagesStringInternal()")
```

The first call to `getMessagesStringInternal()` gives the notificaitons caused by loading model `M`, the second call gives the error caused by calling `checkModel(M)`.
Note. I haven't tried running the python code but I guess it will work.

You can also see how `getMessagesStringInternal()` is used in OMEdit here https://github.com/OpenModelica/OpenModelica/blob/master/OMEdit/OMEditLIB/OMC/OMCProxy.cpp#L611

adeas31:
Let me know once its ready for review.

syntron:
> Let me know once its ready for review.

@adeas31 all done including `getMessagesStringInternal()` in this PR

Please follow the folloing order for review: PR #281, PR #282 (needs a simple conflict solution), PR #278

PR #279 is currently not updated as it would be on top of PR #278

Besides these, there are some smaller cleanups pending

adeas31:
All the dependent PRs are merged. Lets rebase this one now.

OMPython/OMCSession.py
```diff
@@ -325,6 +326,9 @@ def __init__(self, timeout=10.00,
         # connect to the running omc instance using ZMQ
         self._connect_to_omc(timeout)
 
+        self._re_log_entries = None
+        self._re_log_raw = None
```
adeas31, line 330: Do we really need to make them class members?

OMPython/OMCSession.py
```diff
@@ -549,6 +553,62 @@ def sendExpression(self, command, parsed=True):
             return None
         else:
             result = self._omc.recv_string()
+
+            if command == "getErrorString()":
+                # no error handling if 'getErrorString()' is called
+                pass
+            elif command == "getMessagesStringInternal()":
+                # no error handling if 'getMessagesStringInternal()' is called; parsig NOT possible!
```
adeas31, line 561: Typo. Should be `parsing` instead of `parsig`.

OMPython/OMCSession.py
```diff
@@ -549,6 +553,62 @@ def sendExpression(self, command, parsed=True):
...
+                if parsed:
+                    logger.warning("Result of 'getMessagesStringInternal()' cannot be parsed - set parsed to False!")
+                    parsed = False
+            else:
+                # allways check for error
+                self._omc.send_string('getMessagesStringInternal()', flags=zmq.NOBLOCK)
+                error_raw = self._omc.recv_string()
+                # run error handling only if there is something to check
+                if error_raw != "{}\n":
+                    if not self._re_log_entries:
```
adeas31, line 571: `self._re_log_entries` and `self._re_log_raw` can be local variables here.

adeas31 (commented):
LGTM.

OMPython/OMCSession.py
syntron: I defined them as class variables to do the initialisation once (if needed at all). Later, the compiled re struture is just reused ... However, if the expected numbrs of calls to this part is low, it makes sense to run this each time.

adeas31: The actual re compilation is run each time, its just they are initialized to none only once. I don't think initializing them to none locally is expensive.

syntron: in OMCSession.py line 584ff:

    if not self._re_log_entries:
        self._re_log_entries = re.compile(...)
 
    if not self._re_log_raw:
        self._re_log_raw = re.compile(...)

Compilation is done only once (on first call; self._re_* is None); after that, the variables are set (not None) and, thus, the compiled version is used ...

OMPython/OMCSession.py
syntron: fixed in latest update

adeas31 (approved):

PR #281, opened 2025-05-02
[OMCSessionBase.ask()] define opt as list

syntron:
define argument opt for OMCSessionBase._ask() as list - build string out of the content in a controlled way (this was part of the extended version of PR https://github.com/OpenModelica/OMPython/pull/270)

additional changes:
* rename ask() => _ask() to indicate that this is a function internal for the class

PR #282, opened 2025-05-02
Define omc session cmd

syntron:
rename OMCSession base as OMCSessionCmd and restructure it such that it is an independend function which contains only the API calls to OMC (was the final state of PR #270)

based on PR #270, #273, #280 and #281

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -46,7 +46,7 @@ import pathlib
 from dataclasses import dataclass
 from typing import Optional
 
-from OMPython.OMCSession import OMCSessionBase, OMCSessionZMQ
+from OMPython.OMCSession import OMCSessionZMQ
 
 # define logger using the current module name as ID
 logger = logging.getLogger(__name__)
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -117,7 +117,7 @@ class ModelicaSystem:
             variableFilter: Optional[str] = None,
             customBuildDirectory: Optional[str | os.PathLike] = None,
             omhome: Optional[str] = None,
-            session: Optional[OMCSessionBase] = None,
+            session: Optional[OMCSessionZMQ] = None,
             build: Optional[bool] = True
             ):
         """Initialize, load and build a model.
--- a/OMPython/OMCSession.py
+++ b/OMPython/OMCSession.py
@@ -3,6 +3,8 @@
 Definition of an OMC session.
 """
 
+from __future__ import annotations
+
 __license__ = """
  This file is part of OpenModelica.
 
@@ -33,7 +35,6 @@ __license__ = """
 """
 
 import shutil
-import abc
 import getpass
 import logging
 import json
@@ -80,9 +81,12 @@ class OMCSessionException(Exception):
     pass
 
 
-class OMCSessionBase(metaclass=abc.ABCMeta):
+class OMCSessionCmd:
 
-    def __init__(self, readonly=False):
+    def __init__(self, session: OMCSessionZMQ, readonly: Optional[bool] = False):
+        if not isinstance(session, OMCSessionZMQ):
+            raise OMCSessionException("Invalid session definition!")
+        self._session = session
         self._readonly = readonly
         self._omc_cache = {}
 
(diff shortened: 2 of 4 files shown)


adeas31:
Time to rebase this one.

syntron:
@adeas31 PR is rebased

syntron:
Interesting from my look into the code it is the follwoing class dependency:

subprocess.TimeoutExpired
=> SubprocessError
=> Exception
=> BaseException

So the error message is unexpected ...

Furthermore, the following code works:

`
import subprocess

print('START')

cmd = ['sleep', '30s']
seconds = 2

try:
    output = subprocess.run(cmd, stderr=subprocess.STDOUT, timeout=seconds)
except subprocess.TimeoutExpired as ex:
    print(f'Timeout: {ex} {type(ex)}')

print('Done')
`

It results in:

> START
> Timeout: Command '['sleep', '30s']' timed out after 1.9999794609975652 seconds <class 'subprocess.TimeoutExpired'>
> Done

I also checked if wait() raises TimeoutExpired => Yes; see [https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait](url)

Tested with 3.13.3

syntron:
@adeas31 could you retry with the latest commit added?

I think this is not related to the main code but to the handling of the last command you did:

> >>> omc.sendExpression("quit()")

This stops OMC, i.e. _omc is set to None

> >>> quit()

here, the destructor of the class (=> __del__()) is called, which in turn calls self.sendExpression("quit()") and it fails as _omc is no longer defined

I'm still not sure why TimeoutExpired is indicated as the source of the error ...

syntron:
Updated with a more generic approach ...

syntron:
@adeas31 Could it be that this is not really an error but more a python / debugger thing? I just noted that the test script (your code) runs fine if I 'run' it. Even if I stop in line 547: 'self._omc.close()' and after that run continue all is fine. Only if I do a step by step execution it fails ...

Thus, the latest commit is perhaps not needed at all ...

adeas31:
Its still the same.
Definitely the issue is related to handling of `quit()` command.

syntron:
> @adeas31 Could it be that this is not really an error but more a python / debugger thing? I just noted that the test script (your code) runs fine if I 'run' it. Even if I stop in line 547: 'self._omc.close()' and after that run continue all is fine. Only if I do a step by step execution it fails ...
> 
> Thus, the latest commit is perhaps not needed at all ...

Just as Information: I'm using PyCharm and running the code in the debugger of the tool; it returns something like this

> Exception ignored in: <function OMCSessionZMQ.__del__ at 0x000001E7FEB911C0>
> Traceback (most recent call last):
>   File "[...]\OMPython\OMCSession.py", line 330, in __del__
>   File "[...]\OMPython\OMCSession.py", line 525, in sendExpression
>   File "_pydevd_bundle\\pydevd_pep_669_tracing_cython.pyx", line 449, in _pydevd_bundle.pydevd_pep_669_tracing_cython.py_line_callback
>   File "_pydevd_bundle\\pydevd_pep_669_tracing_cython.pyx", line 169, in _pydevd_bundle.pydevd_pep_669_tracing_cython._getframe
> AttributeError: 'NoneType' object has no attribute '_getframe'

Due to this error message, I think it is *not* OMPython related but an error of the 'quit()' command, i.e. the closing of the debugger

adeas31:
Yes, the OMPython code itself is correct. The message shown is expected message. Here's the relevant portion of [the docs](http://docs.python.org/2/reference/datamodel.html#object.__del__):

> Warning: Due to the precarious circumstances under which __del__() methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead. Also, when __del__() is invoked in response to a module being deleted (e.g., when execution of the program is done), other globals referenced by the __del__() method may already have been deleted or in the process of being torn down (e.g. the import machinery shutting down). For this reason, __del__() methods should do the absolute minimum needed to maintain external invariants. Starting with version 1.5, Python guarantees that globals whose name begins with a single underscore are deleted from their module before other globals are deleted; if no other references to such globals exist, this may help in assuring that imported modules are still available at the time when the __del__() method is called.

So what we see is expected.

> Thus, the latest commit is perhaps not needed at all ...

True.

adeas31 (approved):

PR #285, opened 2025-05-08
Adding build as argument to ModelicaSystem constructor

Nmichele29:
### Purpose

Enables the possibility to instanciate a ModelicaSystem object without compiling the Modelica model called, therefore allowing the user to modify values of structural parameters through `mod.sendExpression("setParameterValue(model, parameter, value)", parsed = False)
` and only compiling the model manually with `mod.buildModel()` once the values are modified. This way, we can avoid to compile the model when we call the ModelicaSystem constructor, change structural parameters and have to compile the model again to take the changes in account.

### Approach

A new optional argument `build` is added to the ModelicaSystem constructor, which is a boolean value defaulting to `True`. The `mod.buildModel()` line in constructor is then enclosed in a `if` loop controlled by the value of `build`.

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -221,7 +221,7 @@ class ModelicaSystem:
         elif fileName is None and modelName is not None:
             self.loadLibrary(lmodel=self.lmodel)
         
-        if build :
+        if build:
             self.buildModel(variableFilter)
 
     def setCommandLineOptions(self, commandLineOptions: str):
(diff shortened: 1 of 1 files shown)


Nmichele29:
I'm sorry about these back and forths, am I in the right repositary this time ? 

Also, through a little testing my change does raise an uncontrolled error message if one tries to simulate a model without ever building it (by setting `build` to `False` and never calling `mod.buildModel()`. Should I add some Boolean check to make sure the model is built before simulated and raise some kind of error in the `mod.simulate()` method if not ?

arun3688 (commented):
@Nmichele29 Currently the tests are failining because of trailing whitespaces can you remove all the trailing whitespaces and then commit your code, if you are using vscode, just open the file and press remove trailing whitespaces

Nmichele29:
Ok thanks for the help, I hope it's good now !

arun3688:
@Nmichele29 Thanks for the contribution

PR #292, opened 2025-05-28
[ModelicaSystem] Cleanup & mypy

syntron:
cleanup within ModelicaSystem and mypy updates

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -522,7 +522,7 @@ class ModelicaSystem:
         return self.sendExpression(exp)
 
     def xmlparse(self):
-        if not self.xmlFile.exists():
+        if not self.xmlFile.is_file():
             raise ModelicaSystemError(f"XML file not generated: {self.xmlFile}")
 
         tree = ET.parse(self.xmlFile)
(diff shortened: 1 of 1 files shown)


OMPython/ModelicaSystem.py
```diff
@@ -482,22 +482,22 @@ def setTempDirectory(self, customBuildDirectory) -> pathlib.Path:
...
-        logger.debug("varFilter=%s", varFilter)
+
         buildModelResult = self.requestApi("buildModel", self.modelName, properties=varFilter)
         logger.debug("OM model build result: %s", buildModelResult)
 
         self.xmlFile = pathlib.Path(buildModelResult[0]).parent / buildModelResult[1]
         self.xmlparse()
 
-    def sendExpression(self, expr, parsed=True):
+    def sendExpression(self, expr: str, parsed: bool = True):
```
adeas31, line 500: This should be `parsed: Optional[bool] = True`, right?

syntron: Not in this case as True is a valid bool value; it would be needed in case of `parsed: Optional[bool] = None`

adeas31 (approved):

PR #294, opened 2025-05-28
[ModelicaSystem] reorder input (mypy)

syntron:
Class ModelicaSystem uses several imputs which are all optional; however, modelName must be present.

This PR reorders the inputs such that modelName is the first parameter. This also fixes one of the last mypy warnings.

Please keep in mind, that this is a breaking change - all calls to ModelicaSystem which do not use kwargs will fail!

On top of PR #292

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -314,8 +314,8 @@ class ModelicaSystemCmd:
 class ModelicaSystem:
     def __init__(
             self,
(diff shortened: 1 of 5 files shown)


adeas31 (approved):
> Please keep in mind, that this is a breaking change - all calls to ModelicaSystem which do not use kwargs will fail!

As far as I can see, you updated all the reference calls so we are good to go, right?

adeas31:
Please rebase the PR.

PR #299, opened 2025-06-11
Make getters raise KeyError instead of returning "NotExist"

ondras12345:
This closes https://github.com/OpenModelica/OMPython/issues/260

I chose to use `KeyError` over `ModelicaSystemError`, because it makes the implementation easier (dicts already raise it for us), and because I find it useful to be able to distinguish invalid input from other errors. If you prefer `ModelicaSystemError`, let me know and I'll change it.

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -472,12 +472,12 @@ class ModelicaSystem:
         # create a unique temp directory for each session and build the model in that directory
         if customBuildDirectory is not None:
             if not os.path.exists(customBuildDirectory):
-                raise IOError(customBuildDirectory, " does not exist")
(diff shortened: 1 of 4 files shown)


syntron:
@ondras12345 I tried to use project specific exceptions if possible (and meaningful); thus, on an exception I could catch these to ensure there is no project related error. 

My changes to the code are based on this assumption; i.e. raise ModelicaSystemError in ModelicaSystem.py and OMCSessionException in OMCSession.py if the case is due to OMC or the related Python code.

syntron:
@ondras12345 I'm fine with your approach!

My way is focused on the possibility to catch all errors related to ModelicaSystem by this one exception; this would mean that a check for this would ensure that the functionality of ModelicaSystem is covert.

adeas31 (approved):

PR #309, opened 2025-06-17
[ModelicaSystem] only local omc process allowed

syntron:
ModelicaSystem relies on read access to files and the possibility to execute the compiled model; thus, it *only* works if OMC is started locally

Use OMCProcessLocal to ensure that this requirement is defined

Based on PR #295

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -49,7 +49,7 @@ from typing import Optional, Any
 import warnings
 import xml.etree.ElementTree as ET
 
-from OMPython.OMCSession import OMCSessionException, OMCSessionZMQ
+from OMPython.OMCSession import OMCSessionException, OMCSessionZMQ, OMCProcessLocal
 
 # define logger using the current module name as ID
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


adeas31:
Does this mean we can't use docker and wsl instances with ModelicaSystem? Is this just because if I/O permission issues?

syntron:
> Does this mean we can't use docker and wsl instances with ModelicaSystem? Is this just because if I/O permission issues?

@adeas31 no, the reason is that ModelicaSystem accesses the file system directly without using OMC (create csv file; read result data, etc.). Due to this, the class can only used locally ...

syntron:
Another big issue is the call to the executable in simulate() and linearize() using ModelicaSystemCmd(). for this case also a solution is needed - how to call / run the executable from the host within docker / wsl?

syntron:
I think this is a quite good descission as it allows to run control the execution of the model in more detail; see also the possibility of running DoEs (see PR #312).

However, it also limits the usability of the class ModelicaSystem if it is not run locally, i.e. in docker or WSL. I do not know the history - was ModelicaSystem in the past useable via docker?

The linux case still needs to be tested (including external C libraries like SDF) ... currently I'm only able to run OMC in docker or on Windows.

adeas31:
> However, it also limits the usability of the class ModelicaSystem if it is not run locally, i.e. in docker or WSL. I do not know the history - was ModelicaSystem in the past useable via docker?

@arun3688 can you please comment on this? Is `ModelicaSystem` useable via docker?

Saw that docker test is skipped so I enabled it in PR #319. However, the test is very basic. We should try to enhance it to see if we can test model simulation using docker.

@syntron I am assuming that `ModelicaSystem` is always used locally, @arun3688 will confirm it, so this PR makes sense to restrict it to local process. For now I will review this PR as it is and we will see if we can fix #315 in future.

adeas31 (approved):
LGTM.

Issue #230, opened 2024-11-24
Handle process externally in ModelicaSystem.simulate and ModelicaSystem.linearized


kiroorooki:
Hello! 
I've noticed that we could run simulations in parallel if we handled the process outside the simulate() method.

It is useful to me because I need to run a log of simulations, and running them in parallel saves me a lot of time.
In order to keep the same functionality, I added a boolean handleProcessOutside, and if set to true, I don't wait() and terminate() the process, I simply return it, so the user can handle the process themselves. That way people can use the library the same way, or get extended functionality if needed with just a boolean.

In my case, I run all my simulations, and I make sure all of them are done later when I save my results (opening .mat file, converting things to csv...).

I just realized the library changed quite a bit, but I believe the same functionality can still be implemented in _run_cmd and propagated to the methods using it (simulate and linearized).

Thanks for your hard work, I hope my suggestion makes sense!

This is how I currently implemented it in my version for reference (quite outdated):
`

    def simulate(self, resultfile=None, simflags=None, verbose=True, handleProcessOutside=False):  # 11
        """
        This method simulates model according to the simulation options.
        usage
        >>> simulate()
        >>> simulate(resultfile="a.mat")
        >>> simulate(simflags="-noEventEmit -noRestart -override=e=0.3,g=10) set runtime simulation flags
        """
        if(resultfile is None):
            r=""
            self.resultfile = os.path.join(self.tempdir, self.modelName + "_res.mat").replace("\\", "/")
        else:
            r=" -r=" + resultfile
            self.resultfile = resultfile

        # allow runtime simulation flags from user input
        if(simflags is None):
            simflags=""
        else:
            simflags=" " + simflags

        overrideFile = os.path.join(self.tempdir, '{}.{}'.format(self.modelName + "_override", "txt")).replace("\\", "/")
        if (self.overridevariables or self.simoptionsoverride):
            tmpdict=self.overridevariables.copy()
            tmpdict.update(self.simoptionsoverride)
            # write to override file
            file = open(overrideFile, "w")
            for (key, value) in tmpdict.items():
                name = key + "=" + value + "\n"
                file.write(name)
            file.close()
            override =" -overrideFile=" + overrideFile
        else:
            override =""

        if (self.inputFlag):  # if model has input quantities
            for i in self.inputlist:
                val=self.inputlist[i]
                if(val==None):
[...]

PR #312, opened 2025-06-25
new class ModelicaSystemDoE

syntron:
New class to run DoEs based on the available functionality

Usage (see test_ModelicaSystemDoE):
* define the model
* define all parameters with the values to iterate
* run DoE (simulation of **all** parameter combinations)

based on PR #310 and #311; see also #230


syntron:
It is running but there is still some work need to get it to a working state for all corner cases

OMPython/ModelicaSystem.py
```diff
@@ -1734,3 +1737,373 @@ def getLinearOutputs(self) -> list[str]:
...
+
+
+    if __name__ == "__main__":
+        run_doe()
+    ```
+
+    """
+
+    DICT_RESULT_FILENAME: str = 'result filename'
+    DICT_RESULT_AVAILABLE: str = 'result available'
```
adeas31: Are these actually used?

syntron: One is / one was not; I added these to ensure the same dict keys in the code - variables can be checked but string content not; the new version includes additional commits which update this part and also add some comments

OMPython/ModelicaSystem.py
```diff
@@ -1734,3 +1737,373 @@ def getLinearOutputs(self) -> list[str]:
...
+                sim_param_structure[pk_structure] = pc_structure[idx_structure]
+
+                pk_value = pc_structure[idx_structure]
+                if isinstance(pk_value, str):
+                    expression = f"setParameterValue({self._modelName}, {pk_structure}, \"{pk_value}\")"
+                elif isinstance(pk_value, bool):
+                    pk_value_bool_str = "true" if pk_value else "false"
+                    expression = f"setParameterValue({self._modelName}, {pk_structure}, {pk_value_bool_str});"
+                else:
+                    expression = f"setParameterValue({self._modelName}, {pk_structure}, {pk_value})"
```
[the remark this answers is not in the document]
syntron: Thanks for confirming! I just added a check for double quotes within string variable content ...

syntron:
This currently fails in pytest - rebase order / dependencies; I will solve this (later)!

syntron:
> This currently fails in pytest - rebase order / dependencies; I will solve this (later)!

Working in combination with PR #333 ... not sure how to solve ... merge both into one PR or merge together (this one + #333)

adeas31:
I believe this can be closed now since #333 is merged.

Issue #315, opened 2025-06-27
ModelicaSystem only useable locally - with possible solutions


syntron:
See discussion in PR #309 – ModelicaSystem can only be used locally due to dependencies on access to the file system where OMC is running as well as the need to call the compiled model executable. This limits the usage of the class as it cannot be used with docker or WSL.

There are the following problems to solve (and possible solutions):

1. need of file system interactions => solution: create OMCPath, an implementation of a pathlib.Path object which uses OMC as backend (see detailed analyses which functionality is needed in PR #309)

2. run executable => solution: include a basic functionality within OMCSessionZMQ / OMCProcess* which allows to just(!) call the executable; this would abstract this functionality, i.e. it would prepend the command with ‘docker …’ or ‘wsl …’ as needed (but would also introduce another link / dependency to the system where OMC is running). Here it would be helpful if this could be coded such that it only returns the command to call (as list). Thus, ModelicaSystemCmd would stay independent of the OMCSessionZMQ instance.

3. library search path => solution: see (2); for this item, a solution is needed, even if (1) and (2) are not considered – does models executable using external libraries (example: SDF) run at all using ModelicaSystem with Linux? how to get the needed paths?

@adeas31 feel free to provide your thoughts

syntron:
Please check the following PRs for an implementation:

(1) see PR #317 and #322 
(2) see PR #324
(3) still needs to be tested ... I hope to get an example running using docker ...

Please keep in mind that all of this is only working for Python >= 3.12

syntron:
A solution is available! Please check PR #333 for a branch which contains all changes listed below

PRs:

(1) NO additional dependencies

* 0=>1 (PR #317)
* 0=>1 (PR #329)
* 0=>1 (PR #320)
* 0=>1 (PR #321)
* 0=>1 (PR #330)
* 0=>1 (PR #314)
* 0=>2 (PR #331)
* 0=>2 (PR #332)
* 0=>3 (PR #312)

(2) WITH dependencies

* 1=>2 (PR #322) - (OMCPath, ModelicaSystem_workdir, ModelicaSystem_linearize, ModelicaSystem_xml, ModelicaSystem_rewrite_set_functions, ModelicaSystem_prepare_OMCPath)
* 2=>3 (PR #324) - (ModelicaSystem_use_OMCPath, OMCSession_executable, OMCProcess_pathlib)
[...]

PR #317, opened 2025-06-28
OMCPath

syntron:
new class OMCPath which implements path handling based on OMC commands

currently just a prove of concept with a test case using docker

see issue #315 and PR #309


syntron:
@Sonyoyo could you please test if this is working on WSL and provide soem feedback? I'm just coding blind regarding this option ...

Feel free to use the testcase in `tests.test_OMCPath.test_OMCPath_WSL()` - it is working well on local and docker; thus, should also work on WSL (I hope)

syntron:
@ondras12345 could you please advice on the unittest?

In `tests.test_OMCPath` tests are defined which use different backends (local / docker / WSL). I implemented it by using different functions to define the test and the unique code is in a helper function which is called by each test.

Issue #254, opened 2025-04-15
next release?


syntron:
Is a new release of the OMPython package planned? The last release mentioned on github is v3.5.2 (2024/06/24) and 3.6.0 is available via pip (2024/07/12). However, the later version cannot be found on github; there is also no tag for this version.

syntron:
@adeas31: sounds good - no problem to wait

I have some additional changes in the pipeline - mainly some cleanup of the files:

* move classes from __init__.py into separate files (=> OMCSession; ModelicaSystem)
* move OMParser/__init__.py into just a directory

Reason:
* separate the differnet classes / parser such that specific commits touch only the corresponding file
* use __init__.py only for the package related definitions (public interface of the package)

I will check if I can prepare a PR for this

ondras12345:
I also have some changes ready that are waiting for https://github.com/OpenModelica/OMPython/pull/249 to be merged (otherwise there will be conflicts):
- Fix two spaces in simflags causing error
- Do not ignore subprocess returncode

I would also like to look into refactoring `createCSVData` and cleaning up the docstrings, but I'm not sure if I'll be able to do that in time for the release.

syntron:
@adeas31 I have two PRs open which are ready for merge: PR #294 (just updated) and PR #297 (new; small cleanups)

PR #295 is something for later as I have to build up the capability to test OpenModelica in docker first; the non-docker part is working well ;-)

ondras12345:
I don't have anything ready.
https://github.com/OpenModelica/OMPython/pull/289 is just a tool for developers, so we don't care about releases.
I would like to tackle https://github.com/OpenModelica/OMPython/issues/260 at some point, but that can easily wait for a later release. I'm currently busy with university exams.

arun3688:
@adeas31 Yes we can make `4.0.0` and that looks good

ondras12345:
Another idea: currently, there we are exposing internal functions that probably shouldn't be part of the public API. For example, I see no reason why the user would need to call `ModelicaSystem.xmlparse()` directly. Maybe these functions should be renamed to start with an underscore (e.g. `_xmlparse()`). While this does not prevent outside users from calling the function, it is a generally understood to mean that [the function should only be used internally](https://peps.python.org/pep-0008/#descriptive-naming-styles).

syntron:
@ondras12345 Besides functions this would also help to define the scope for variables

syntron:
[...]

PR #321, opened 2025-07-11
ModelicaSystem - remove xml_file as class variable

syntron:
cleanup handling of the xml file & remove another class variable

preparation for OMCPath (PR #317)

needs also PR #321

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -1558,10 +1558,10 @@ class ModelicaSystem:
         overrideLinearFile = self._tempdir / f'{self._model_name}_override_linear.txt'
 
         with open(file=overrideLinearFile, mode="w", encoding="utf-8") as fh:
-            for key, value in self._override_variables.items():
-                fh.write(f"{key}={value}\n")
-            for key, value in self._linearization_options.items():
-                fh.write(f"{key}={value}\n")
+            for key1, value1 in self._override_variables.items():
+                fh.write(f"{key1}={value1}\n")
+            for key2, value2 in self._linearization_options.items():
+                fh.write(f"{key2}={value2}\n")
 
         om_cmd.arg_set(key="overrideFile", val=overrideLinearFile.as_posix())
 
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -480,8 +480,10 @@ class ModelicaSystem:
         buildModelResult = self._requestApi("buildModel", self._model_name, properties=varFilter)
         logger.debug("OM model build result: %s", buildModelResult)
 
-        self._xml_file = pathlib.Path(buildModelResult[0]).parent / buildModelResult[1]
-        self._xmlparse()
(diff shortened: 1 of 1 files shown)


syntron:
@ondras12345 do you have an idea why the last commit is needed to make mypy happy?

on my local checks there was no problem; I do not understand the background of the error in line 1528: `n` is str but should be int - why is `n` a str? There was no change in this code ...

ondras12345:
`mypy` does not like it when you reuse variable names with different types.
https://github.com/OpenModelica/OMPython/blob/f958964b754a496ef4042006133785999d7f38f8/OMPython/ModelicaSystem.py#L1481
Here, `n` is a str. Its scope is not limited to that for loop, so it still exists when you get here:
https://github.com/OpenModelica/OMPython/blob/f958964b754a496ef4042006133785999d7f38f8/OMPython/ModelicaSystem.py#L1523-L1528
`mypy` does not know the type of `result`, so it assumes it is compatible with the previous (implicit) type of `n`. And when you try to pass that to `LinearizationResult`, it gives you the
```
OMPython/ModelicaSystem.py:1527: error: Argument 1 to "LinearizationResult" has incompatible type "str"; expected "int"  [arg-type]
```
This is my best guess.

It looks like there is now an [option in mypy to allow variable redefinition](https://github.com/python/mypy/pull/6197), but I'm not sure if it would help in this case. If you want to try it, add this to `pyproject.toml`:
```
[tool.mypy]
allow_redefinition = true
```

ondras12345:
> However, did you notice, that current master ([3e9c55b](https://github.com/OpenModelica/OMPython/commit/3e9c55ba84298a973e5e9dbc6872c0917c4cb9a5)) does pass mypy without error and commit [f958964](https://github.com/OpenModelica/OMPython/commit/f958964b754a496ef4042006133785999d7f38f8) from this PR fails? There are no changes in this part of the code!

No idea why this is happening. We specify an exact version of mypy, so it shouldn't be that the new CI run pulled a newer version:
https://github.com/OpenModelica/OMPython/blob/3e9c55ba84298a973e5e9dbc6872c0917c4cb9a5/.pre-commit-config.yaml#L29-L30
And we use
https://github.com/OpenModelica/OMPython/blob/3e9c55ba84298a973e5e9dbc6872c0917c4cb9a5/.github/workflows/Test.yml#L38-L39
so it should check all files in each run.

syntron:
@adeas31 Sorry for the many PRs - I was focused on getting ModelicSystem running in docker and WSL that the other stuff was moved aside. This is currently done and defined in issue #315. Besides that, some smaller (independent) changes exist which are in my fork but do not have at PR at the moment. I will add these also to the issue #315 such that it gives an overview. 

Regarding milestones, I will indicate it also at this place. From my side, the release could be done now. The OMCPath / ModelicaSystemDoE stuff could then be defined in a 4.1.0 release. I will check all smaller changes if it would makes sense to include them in a 4.0.0 release and mark them accordingly.

See also issue #254

adeas31:
Great! I like all the work you are doing.
Please add milestone for all open PRs so manage them accordingly.

adeas31 (approved):

PR #324, opened 2025-07-12
ModelicaSystemCMD - use OMCPath

syntron:
Use OMCPath in ModelicaSystemCMD - this allows to run ModelicaSystem simulations using docker or WSL

Please check `tests.test_ModelicaSystem.test_getSolutions_docker` for an example

Based on PR #322


syntron:
@Sonyoyo feel free to also test this PR - it should allow you to use ModelicaSystem in combination with OMC in WSL

Test case to check:  `tests.test_ModelicaSystem.test_getSolutions_docker` - this can be easily adapted to run on WSL ...

Please keep in mind that Python >= 3.12 is needed!

PR #326, opened 2025-08-01
Better interface for `setParameters`

SengerM:
### Purpose

```python
setParameters(Name=value)
setParameters(Name1=value1, Name2=value2)
``` 
are more clean and natural than 
```python
setParameters("Name=value")
setParameters(["Name1=value1","Name2=value2"])
```
It should be backwards compatible, meaning that it still accepts the strings.

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -1175,15 +1175,22 @@ class ModelicaSystem:
         """
         return self._setMethodHelper(cvals, self._continuous, "continuous", self._override_variables)
 
-    def setParameters(self, pvals):  # 14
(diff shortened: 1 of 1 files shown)


syntron:
@SengerM would this work on top of PR #314? it creates the option to use dictionaries for all set*() functions ...

syntron:
I see the dict based interface as preferable if lots of options are handled which are defined before as dictionary. My idea is to add an additional function which just translates the argument key based variant to the dict based version.

> I would also only keep the current interface with strings for backwards compatibility, and add a warning that it will be deprecated in the future.

This is done in PR #314; I also prepared a commit which would remove the depreciated / old way of setting the options.

PR #327, opened 2025-08-01
`RLC_series` example

SengerM:
### Purpose

Help new users to get started.

### Approach

Add an `examples` folder with examples with different complexities, ready to be run.


syntron:
Would it make sense to add a github workflow to test these examples?

This would mean, that each example needs a defined entry function with a return value on success (i.e. have each example check if there are the expected results and/or plots are created). Then, this could be checked similar to the unittests ...

The advantage is, that any change breaking the example(s) can easily be found

PR #345, opened 2025-08-15
[ModelicaSystem] improve set functions

syntron:
improve input handling for set*() functions
    
* use a Pythonic way for input:
```
    setParameters(a=123)

    param = {'a': 123}
    setParameters(**param)
```
Based on PR #314 with input from SengerM (see PR #326); this replaces PR #326

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -1115,7 +1115,8 @@ class ModelicaSystem:
 
     @staticmethod
     def _prepare_input_data(
-            raw_input: str | list[str] | dict[str, Any],
+            input_args: Any,
+            input_kwargs: dict[str, Any],
     ) -> dict[str, str]:
         """
         Convert raw input to a structured dictionary {'key1': 'value1', 'key2': 'value2'}.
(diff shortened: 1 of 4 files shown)


syntron:
@adeas31 one thing I did not checked but which is quite important to confirm here: are there the same rules for variable names in Python and OpenModelica? I assume yes and then all is fine - if not this PR should not be applied!

adeas31:
> @adeas31 one thing I did not checked but which is quite important to confirm here: are there the same rules for variable names in Python and OpenModelica? I assume yes and then all is fine - if not this PR should not be applied!

Which variables are you talking about? Did you mean Modelica and not OpenModelica?

adeas31:
Also as I understand this supersedes PR #326.

SengerM:
> Which variables are you talking about? Did you mean Modelica and not OpenModelica?

[The rules for identifiers in Python](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) and [the ones for Modelica](https://specification.modelica.org/maint/3.6/lexical-structure.html#S3.SS1.p1).

> The reason for this is that by using the code in this PR, the Modelica variables are used as Python variables. Thus, the rules should match in this direction.

Even if the naming rules are a bit different, you can overcome this issue if the current string interface is maintained for those specific cases.

Issue #144, opened 2021-08-16
ModelicaSystem does not show plot window


bilderbuchi:
I am trying to plot results using the ModelicaSystem API, but the plot window is not shown. 
This is on Windows, using OMPython master and a recent OM nightly(OpenModelica v1.18.0-dev.beta1 (64-bit)).

Based on examples from the documentation, using the omc API works, it shows a plot window (the sleep is just there so that the window does not immediately disappear when the script is finished):
```
from OMPython import OMCSessionZMQ
omc = OMCSessionZMQ()
cmds = [
  'loadFile(getInstallationDirectoryPath() + "/share/doc/omc/testmodels/BouncingBall.mo")',
  "simulate(BouncingBall)",
  "plot(h)"
  ]
for cmd in cmds:
  answer = omc.sendExpression(cmd)
  print("\n{}:\n{}".format(cmd, answer))
from time import sleep
sleep(5)
```

Doing the same thing (I hope?!) with the ModelicaSystem API does **not** show a window:
```
from time import sleep

from OMPython import OMCSessionZMQ
omc = OMCSessionZMQ()
model_path=omc.sendExpression("getInstallationDirectoryPath()") + "/share/doc/omc/testmodels/"
from OMPython import ModelicaSystem
mod=ModelicaSystem(model_path + "BouncingBall.mo","BouncingBall")
mod.simulate()
print('now plot ')
mod.sendExpression('plot(h)')
print(mod.requestApi('getErrorString'))
print('done, now sleep')
sleep(5)
```
output (notice: nothing in the error string):
```
LOG_SUCCESS       | info    | The initialization finished successfully without homotopy method.
LOG_SUCCESS       | info    | The simulation finished successfully.
now plot

done, now sleep
```
I tried to find out what's going wrong, but couldn't.

arun3688:
@bilderbuchi the `ModelicaSystem()` API uses a different approach to produce the simulation results, it basically builds the model first and then` simulate()`  API uses the simulation executable to produce the result file and the result file is not stored in`sendExpression()` when using `ModelicaSystem()`  API, so inorder to use the `plot() ` API with `ModelicaSystem()` we have to explicitly specify the result file to the `plot() ` API  some thing like below


```
mod=ModelicaSystem(model_path + "BouncingBall.mo","BouncingBall")
mod.simulate()
print('now plot ')
mod.resultfile ### this will show you the resultfile location
[...]

PR #352, opened 2025-08-26
 [ModelicaSystem] add plot() function

syntron:
add a plot() function; see issue #144

just because it can be done ...; for 4.1.0

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -38,15 +38,12 @@ import logging
 import numbers
 import numpy as np
 import os
-import platform
-import re
-import subprocess
 import textwrap
 from typing import Optional, Any
 import warnings
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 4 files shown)


OMPython/ModelicaSystem.py
```diff
@@ -1132,6 +1132,36 @@ def simulate(
...
+            plotdata: str,
+            resultfile: Optional[str | os.PathLike] = None,
+    ) -> None:
+        """
+        Plot a variable using OMC; this will work for local OMC usage only (OMCProcessLocal).
+        """
+
+        if not isinstance(self._getconn.omc_process, OMCProcessLocal):
+            raise ModelicaSystemError("Plot is using the OMC plot functionality; "
+                                      "thus, it is only working if OMC is running locally!")
```
adeas31: Is it still for local omc even after adding OMCPath class?

PR #367, opened 2025-11-10
[ModelicaSystem] add docstring

syntron:
add missing docstring for class ModelicaSystem

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -322,6 +322,10 @@ class ModelicaSystemCmd:
 
 
 class ModelicaSystem:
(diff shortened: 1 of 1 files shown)


OMPython/ModelicaSystem.py
```diff
@@ -322,6 +322,10 @@ def parse_simflags(simflags: str) -> dict[str, Optional[str | dict[str, Any] | n
 
 
 class ModelicaSystem:
+    """
+    Class to run OpenModelica simulations.
```
adeas31: Perhaps a better docstring is `Class to simulate a Modelica model.`

PR #371, opened 2025-11-10
test cleanup

syntron:
simplify definition of unittests


syntron:
ups; needs PR #365 ...

PR #404, opened 2026-01-11
RFC - definition of a runner class using as much of the existing code as possible

syntron:
Content (lots of changes / these will be splitted into several PRs if the general approach is approved):

* OMCSession.py: rename real OMCPath* classes as privat (start with '_')
* OMCSession: define OMCPathDummy and OMCSessionDummy => no OMC server
* ModelExecution.py: include all code related to model execution (from OMCsession*; ModelicaSystemCmd)
* ModelicaSystemDoE.py: seperate file for this class
* ModelicaSystemDoE.py: use ModelicaSystem / reduce number of arguments to __init__
* ModelicaSystem(Base).py: split getOutputs() and getContinuous() to Initial and Final data
* ModelicaSystem(Base).py: getOutputs*() / getContinuous*() - always return np.float64
* ModelicaSystemBase.py: define a basic version of ModelicaSystem
* ModelicaSystemRunner.py: definition of a 'runner' class using ModelicaSystemBasic / OMCSessionDummy (see PR #401)
* test_ModelicaSystemRunner.py: test case for ModelicaSystemRunner
* test_*: update of test cases based on above listed changes

Compared to PR #401 this solution:
* still uses the existing OMPython requirements (dependencies of the Runner class; i.e. numpy, etc.)
* shares nearly all code with the current definitions (by using OMCPathDummy and OMCSessionDummy)
* uses `|` instead of `Union`

Further changes could do additional modifications such that the 3rd party packages (ZMQ) are not needed at all in the runner code path.

All tests are running without error / tested in https://github.com/syntron/OMPython/actions/runs/20865402893

Please comment!


arun3688 (approved):
looks good

syntron:
@joewa any feedback from your side?

@Sonyoyo could you test the runner functionality for WSL? I do not have this option available to me ...

adeas31:
> @adeas31 I'm working with smaller commits which build up the PR in serveral steps; however, if these are merged into one commit, were is a high risk for merge conflicts for later PRs especially if these build on top of each other. Would it be OK if I create single step PRs for section B to E (see [#404 (comment)](https://github.com/OpenModelica/OMPython/pull/404#issuecomment-3922322454))? I would keep the details locally but only push one-commit PRs ...

Yeah sounds good.

syntron:
Updated list of open PRs based on this RFC:

* PR #420 - (A010) Use function keyword arguments 
* PR #421 - (A011) update docstring of __init__.py - do not promote depreciated OMCSessionZMQ 
* PR #422 - (A012) reorder imports in __init__.py 
* PR #423 - (A013) [ModelicaSystemDoE] simplify definition 
* PR #424 - (A014) [ModelExecution] update code for execution of a compiled model 
* PR #425 - (A015) [ModelicaSystem] check model executable 
* PR #426 - B001 ModelicaSystem split 
* PR #427 - B002 ModelicaSystemDoE split 
* PR #428 - B003 OmcSession / OMPathABC 
* PR #429 - B004 OMCSession / OMSessionABC 
* PR #430 - C001 Runner definitions 
* PR #431 - D001 OMSession - small fixes 
* PR #432 - D002 OMCSessioZMQ move 
* PR #433 - D003 OMCPath improvements 
* PR #434 - D004 OMSessionRunner 
* PR #435 - D005 OMCSession - classes update 
* PR #436 - D006 ModelicaSystem - small fixes 
* PR #437 - D008 v400 compatibility layer 
* PR #438 - D007 define unit test for v4.0.0 compatibility layer 
* PR #439 - E001 update tests 
* PR #440 - E002 prepare restructure 
* PR #441 - E003a restructure 
* PR #442 - E003b restructure 
* PR #443 - E003c restructure 
* PR #444 - F001 restructure cleanup 

=> all are single comit PRs; if needed, the detailed version can be found at https://github.com/syntron/OMPython/
=> the final state (F001) is matching this PR

* G & H definitions are to come ...

syntron:
> In the next step I'd like the simulation to read timeseries from a file. So the command line has to have `" -csvInput="+self.csvFile` (works for models built with omc 1.25) I cannot see how to make it happen with the updated OMPython (worked for me by setting inputFlag=True before).

Please have a look in PR #447 - is it the code you are searching? (please keep in mind this is untested!)

The code for 1.25 is not working anymore as there is some internal handling of the data to allow to use something like docker or WSL as backend. Thus, the CSV file has to be read and the data merge to the input data used for the model simulation.

joewa:
> Please have a look in PR #447 - is it the code you are searching? (please keep in mind this is untested!)

This is actually not what I wanted. I'm using `-csvInput=myinputtimeseries.csv` only, because the compiled models digest timeseries data in CSV easily. I would be even more happy if some other formats would be supported by the model executables in the future.

I'm calling the simulation like this:
```python
mod.simulate(
            resultfile=resfilepathname,
            simargs={'csvInput': mod_csvFile},  # Old: simflags=f"-csvInput={mod_csvFile}",
            # overrideaux='variableFilter="'+'|'.join(list(res_vars))+'"'  # TODO: how to do this with ModelicaSystemRunner?
        )
```
but this was not working. The problem is that the simarg "csvInput" is overwritten in simulate_cmd (in modelica_system_abc.py)

So I added a check if "csvInput" was already provided in the simargs by the user:
```python
        # always define the result file to use
        om_cmd.arg_set(key="r", val=result_file.as_posix())

        if simargs:
            om_cmd.args_set(args=simargs)

        ...

        if self._inputs:  # if model has input quantities
            for key, val in self._inputs.items():
                ...

            if "csvInput" not in simargs:  # NEW: csvInput provided by user
                # csvfile is based on name used for result file
                csvfile = result_file.parent / f"{result_file.stem}.csv"
                # write csv file and store the name
                csvfile = self._createCSVData(csvfile=csvfile)
    
                om_cmd.arg_set(key="csvInput", val=csvfile.as_posix())

        return om_cmd
```
Does this make sense for you @syntron ?

syntron:
can we continue the discussion about CSV handling in PR #447 - I think it is the right place ...

joewa:
It seems variable_filter is not implemented for ModelicaSystemRunner:
```python
def instantiatemodel404(modelName):
    mod = ModelicaSystemRunner(work_directory='.')
    mod.setup(model_name=modelName, variable_filter=...)
    return mod
```
Is that correct or am I missing something? I think it would be definitely useful for ModelicaSystemRunner too.

syntron:
> It seems variable_filter is not implemented for ModelicaSystemRunner:
> 
> ```python
> def instantiatemodel404(modelName):
>     mod = ModelicaSystemRunner(work_directory='.')
>     mod.setup(model_name=modelName, variable_filter=...)
>     return mod
> ```
> 
> Is that correct or am I missing something? I think it would be definitely useful for ModelicaSystemRunner too.

Quite right - however, it is used in `ModelicaSystemOMC.buildModel()` - this is never called in the runner ... How should it be used? As it is now, this should be removed ...

I expect it has an efect on the xml file created by `buildModel()`; thus, these modification should be done in the provided (defined) xml file, or?

joewa:
Up to version 1.25, it works via `-override` using `-variableFilter` from the command line or via `_override.txt` (cannot test 1.26+ yet). This is an example of a valid `*_override.txt` file:
```
startTime=0
stopTime=24.0
stepSize=0.2
tolerance=1e-06
solver=dassl
outputFormat=mat
variableFilter="dcpm.wMechanical|constantVoltage.i|dcpm.brush.i|torque.tau|hBridge.ref"
```
Assuming that this still works with executables created with 1.26+, I would propose that `ModelicaSystemRunner` shall write this to the `_override.txt` file when `variable_filter` is defined when calling `mod.setup(model_name=modelName, variable_filter=...)`.

This should make the behavior consistent with `buildModel` and we can leave the `_init.xml` file untouched.

In my current workflow, I don't define a `variable_filter` when building. I pick what is needed when calling the runner.

What do you think? @syntron

syntron:
> What do you think? @syntron

@joewa could you test PR #464? It is not directly based on current master but should be good enough to test it.

Please run any follow-up discussion in the PR topic.

PR #435, opened 2026-02-27
D005 OMCSession - classes update

syntron:
(D005) update classes in OMCSession
    
    [OMCSessionPort] fix timeout handling
    
    [OMCSessionDocker*] improve data handling
    
    * move more code to OMCSessionDockerHelper
    * use _docker_omc_start() to differentiate classes
    * define cmd_prefix
    
    [OMCSessionWSL] define cmd_prefix
    
    [OMCSessionWSL] layout fix

### DIFF
--- a/OMPython/OMCSession.py
+++ b/OMPython/OMCSession.py
@@ -20,7 +20,7 @@ import subprocess
 import sys
 import tempfile
 import time
-from typing import Any, Optional, Tuple
+from typing import Any, Optional, Tuple, Type
 import uuid
 import warnings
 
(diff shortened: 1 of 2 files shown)


OMPython/OMCSession.py
```diff
@@ -1596,7 +1597,9 @@ class OMCSessionDockerABC(OMCSessionABC, metaclass=abc.ABCMeta):
 
     def __init__(
             self,
-            timeout: float = 10.00,
+            timeout: float = 10.0,
```
adeas31: Do we need this default in both classes?

syntron: It is defined in each of the Session classes as it is not clear which one will be the entry point - updated in the additional commits: only one definition of the default value in OMSessionABC

adeas31 (approved):

PR #438, opened 2026-02-27
D007 define unit test for v4.0.0 compatibility layer

syntron:
this adds a second unittest branch (test_v400) to check for v4.0.0 compatibility

should be merged *after* D008! (PR #437)


adeas31 (commented):
I don't understand why we need new workflows to test v4.0.0?
What is the purpose of `Test_v400.yml`, `Test_v400_py310.yml` and `Test_v4xx.yml`?


We can test bu just adding an extra step,

```
- name: Run pytest
        uses: pavelzw/pytest-action@v2
        with:
          verbose: true
          emoji: true
          job-summary: true
          custom-arguments: '-v ./tests'
          click-to-expand: true
          report-title: 'Test Report'

- name: Run pytest for v4.0.0
        uses: pavelzw/pytest-action@v2
        with:
          verbose: true
          emoji: true
          job-summary: true
          custom-arguments: '-v ./tests_v400'
          click-to-expand: true
          report-title: 'Test Report'
```

syntron:
> I don't understand why we need new workflows to test v4.0.0? What is the purpose of `Test_v400.yml`, `Test_v400_py310.yml` and `Test_v4xx.yml`?
> 
> We can test bu just adding an extra step,
> 
> ```
> - name: Run pytest
> [...]
> - name: Run pytest for v4.0.0
> [...]
> ```

Quite right; I tested it for myself using the different workflows to activate them as needed with corresponding settings. Would you prefere an update according to this proposal?

syntron:
updated PR - however, this will increase the time to run the unittests (factor 2?)

adeas31:
> > I don't understand why we need new workflows to test v4.0.0? What is the purpose of `Test_v400.yml`, `Test_v400_py310.yml` and `Test_v4xx.yml`?
> > We can test bu just adding an extra step,
> > ```
> > - name: Run pytest
> > [...]
> > - name: Run pytest for v4.0.0
> > [...]
> > ```
> 
> Quite right; I tested it for myself using the different workflows to activate them as needed with corresponding settings. Would you prefere an update according to this proposal?

Yes please. We might remove v4.0.0 tests in future.

syntron:
It fails here but should work on top of latest master (see also test in my fork: https://github.com/syntron/OMPython/actions/runs/25456426602) - updated with the fix included

.github/workflows/Test.yml
```diff
@@ -73,7 +73,17 @@ jobs:
...
+
+      - name: Run pytest based on v4.0.0
+        uses: pavelzw/pytest-action@v2
+        with:
+          verbose: true
+          emoji: true
+          job-summary: true
+          custom-arguments: '-v ./tests_v400'
           click-to-expand: true
           report-title: 'Test Report'
```
adeas31: Update the title so we know this test report belongs to v4.0.0 tests.

adeas31 (approved):

adeas31:
> It fails here but should work on top of latest master (see also test in my fork: https://github.com/syntron/OMPython/actions/runs/25456426602) - updated with the fix included

I don't get it. Why does it timeout here?
So your suggestion is to merge it even if this test fails?
This is already rebased on the master so I wonder how does it work on top of master.

syntron:
> > It fails here but should work on top of latest master (see also test in my fork: https://github.com/syntron/OMPython/actions/runs/25456426602) - updated with the fix included
> 
> I don't get it. Why does it timeout here? So your suggestion is to merge it even if this test fails? This is already rebased on the master so I wonder how does it work on top of master.

Sorry; my wrong in the comment ...

(1) running the unittests twice (master+ v4.0.0) will increase time needed and , thus, the timeout could come
(2) in the first version I used a wrong name for the test-directory and therefore, the test in this PR failed
(3) the test in this PR will fail as it needs changes which are not in this branch but in master

I fixed the bug (see 2) and will not update the title to make it more specific

.github/workflows/Test.yml
syntron: update name & title

adeas31 (approved):

PR #441, opened 2026-02-27
E003a restructure

syntron:
restructure - part1: collect all ModelExecution classes in one file


adeas31:
Please rebase.

OMPython/model_execution.py
```diff
@@ -0,0 +1,358 @@
...
+                        raise ModelExecutionException(f"Invalid value for '-override': {override}")
+                    if kv[0]:
+                        try:
+                            override_dict[kv[0]] = kv[1]
+                        except (KeyError, IndexError) as ex:
+                            raise ModelExecutionException(f"Invalid value for '-override': {override}") from ex
+
+                simargs[parts[0]] = override_dict
+
+        return simargs
```
adeas31, line 358: The classes names sounds a bit confusing to me.

**`ModelExecutionData`** (currently a `dataclass`) actually *does things* — it has `get_cmd()` and `run()` methods. It's the active executor.

**`ModelExecutionCmd`** (currently a regular class) is purely a *builder/configurator* — it accumulates arguments via `arg_set()`, `args_set()`, `arg_get()`, and ultimately produces a `ModelExecutionData` instance via `definition()`. It holds no executable logic itself.

- Something called `...Data` should be a passive data holder (which is exactly what a `@dataclass` normally is), and
- Something called `...Cmd` suggests an active command runner.

But right now it's the opposite: the `@dataclass` runs the process, and the class called `Cmd` just builds up configuration.

If you want to make the roles clearer without changing much, something like `ModelExecutionConfig` (for the builder that accumulates args) and `ModelExecutionRun` (for the dataclass that actually executes) would remove the ambiguity entirely rather than relying on `Cmd` vs `Data` to carry the distinction.

What do you think?

syntron: I'm fine with the renames; would it be OK to do this after the restructure (E003a to E003c)? Else it will quite some work on modifying the data again and again ... I could create a PR for this on top of E003c or F001.

The restructure just move classes and adapt imports, etc. - no change to the content of any class!

Background (as you stated; via a long way of development):
* the `dataclass` collects all the data and has the (minimal) function to execute it
* the other class is just responsible to build / create all the needed variables of the `dataclass`

adeas31: Sure, I will merge this and you can put the change on top.

adeas31 (approved):

PR #446, opened 2026-03-06
(I001) check om version for the model executable

syntron:
due to changes in OMC, there are 2 different ways to define simulation parameters (pre and post 1.26.0). Currently, the version string needs to be correctly defined for the runner class. This PR adds a check of the xml file to extract the version string from there.

see discussion in PR #401


adeas31 (approved):

adeas31:
Please make the linter happy.

syntron:
> Please make the linter happy.

I did focus on PR #468 and #448 as these would be the next; however, to speed it up, we could also jump ...

adeas31 (approved):

Issue #353, opened 2025-08-27
ModelicaSystem.simulate with csv for inputs


FaBiasch:
Hi,

I have the impression that a lot of progress is made with the `ModelicaSystem` module right now, which is great!

I have a question for the current implementation:
If is use 
```
mod = ModelicaSystem(model_path, model_name)
simargs = {'csvInput': input_file_path}
mod.simulate(resultfile=rslt_file_path, simargs=simargs)
```
the input csv is always overridden [here](https://github.com/OpenModelica/OMPython/blob/63139e9e8461f6fc238c207258211eb6738c924e/OMPython/ModelicaSystem.py#L1053).
What is the proposed way to set an input via a csv file?

Thx.

syntron:
@FaBiasch would it be possible for you to set the inputs (= content of the csv file) via the `setInputs()` method?

Else it  could help to add a wrapper which uses a csv file and defines the content in `self._inputs` ...

PR #447, opened 2026-03-06
(I002) setInputsCSV()

syntron:
Read content from a CSV file and use it to define the time based input data.

See issue #353 - currently completly untested from my side!

@FaBiasch could you please tests?

### DIFF
--- a/OMPython/ModelicaSystem.py
+++ b/OMPython/ModelicaSystem.py
@@ -77,14 +77,62 @@ class ModelicaSystem(ModelicaSystemOMC):
             pkey: str,
             args: Any,
             kwargs: dict[str, Any],
-    ) -> Any:
-        param = None
+    ) -> dict[str, Any]:
+        input_args = []
         if len(args) == 1:
-            param = args[0]
-        if param is None and pkey in kwargs:
-            param = kwargs[pkey]
+            input_args.append(args[0])
+        elif pkey in kwargs:
+            input_args.append(kwargs[pkey])
 
-        return param
+        # the code below is based on _prepare_input_data2()
+
+        def prepare_str(str_in: str) -> dict[str, str]:
+            str_in = str_in.replace(" ", "")
+            key_val_list: list[str] = str_in.split("=")
+            if len(key_val_list) != 2:
+                raise ModelicaSystemError(f"Invalid 'key=value' pair: {str_in}")
+            if len(key_val_list[0]) == 0:
+                raise ModelicaSystemError(f"Empty key: {str_in}")
+
+            input_data_from_str: dict[str, str] = {str(key_val_list[0]): str(key_val_list[1])}
+
+            return input_data_from_str
+
+        input_data: dict[str, str] = {}
+
+        if input_args is None:
+            return input_data
+
+        for input_arg in input_args:
+            if isinstance(input_arg, str):
+                warnings.warn(message="The definition of values to set should use a dictionary, "
+                                      "i.e. {'key1': 'val1', 'key2': 'val2', ...}. Please convert all cases which "
+                                      "use a string ('key=val') or list ['key1=val1', 'key2=val2', ...]",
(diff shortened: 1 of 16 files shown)


syntron:
from PR #404:

> > But what against the following code:
> > ```
> >     mod = ModelicaSystemRunner(work_directory='.')
> >     mod.setup(model_name=modelName)
> >     mod.setInputsCSV(csvfile=csvfile)
> >     mod.simulate(resultfile=resfilepathname)
> > ```
> 
> So this creates a CSV file (as input for the model executable) from a CSV file (for `setInputsCSV`) that the user needs to provide. I would try to avoid this workflow if possible.
> 
> * it adds unnecessary overhead (which may become relevant when calling `ModelicaSystemRunner` in huge batches)
> * we need means to customize the location of the intermediate CSV file and users may need to take care about this.
> * users will end up checking two input files in case some things go wrong
> * users will need to take care for additional cleaning & deleting input files
> 
> So I would currently prefer setting `csvInput` in `simargs` as you proposed above @syntron . At least for `ModelicaSystemRunner` there shouldn't be a problem with WSL, Docker, k8s & co.
> 
> On the other hand, I also like the idea of having a simple pythonic interface for timeseries inputs in OMPython. What do you think about
> 
> * `mod.setInputsSeries(series=dict or list of dicts)`. E.g. series=
>   `[{'time': 0.0, 'col1': 1, 'col2': 0.5}, {'time': 0.25, 'col1': 2, 'col2': 0.75}, {'time': 0.5, 'col1': 2.1, 'col2': 0.8}] `
> * or `mod.setInputsDataframe(dataframe=dataframe)`
> 
> I think this would fit the current workflow of OMPython well. And in case the interfaces to the model executables is changed again in a future version, the users wouldn't need to worry about it.

I would not do it via a pandas dataframe as this would add the dependency to OMPython; but using the dictionary would work - at the end, the current code just converts the CSV file to a dictionary to apply it to the existing data.

@joewa would this work for you?

joewa:
Yes, that would work for me @syntron .

When using tables / DataFrames this helper function proved to be useful too:

```python
def toInputs(data: dict[str, list[float]]) -> dict[str, list[tuple[float, float]]]:
    """
    Converts a dictionary of lists (from pandas DataFrame.to_dict(orient='list'))
    into the OMPython setInputs input format.

    Example: mod.setInputs(**toInputs(pdf.to_dict(orient='list')))
    
    Assumes the dictionary contains a key named 'time'.
    """
    if "time" not in data:
        raise ValueError("The provided data must contain a 'time' key.")

    time_series = data["time"]

    inputs = {
        var_name: list(zip(time_series, values))
        for var_name, values in data.items()
        if var_name != "time"
    }

    return inputs
```
Shall we add it as well?

syntron:
> Yes, that would work for me @syntron .
> 
> When using tables / DataFrames this helper function proved to be useful too:
> 
> ```python
> def toInputs(data: dict[str, list[float]]) -> dict[str, list[tuple[float, float]]]:
>     """
>     Converts a dictionary of lists (from pandas DataFrame.to_dict(orient='list'))
>     into the OMPython setInputs input format.
> 
>     Example: mod.setInputs(**toInputs(pdf.to_dict(orient='list')))
>     
>     Assumes the dictionary contains a key named 'time'.
>     """
>     if "time" not in data:
>         raise ValueError("The provided data must contain a 'time' key.")
> 
>     time_series = data["time"]
> 
>     inputs = {
>         var_name: list(zip(time_series, values))
>         for var_name, values in data.items()
>         if var_name != "time"
>     }
> 
>     return inputs
> ```
> 
> Shall we add it as well?

Looks good to me! This can be added as a `staticmethod` to `ModelicaSystemABC`. The check for numeric values is done in setInputs() ...

joewa:
Perfect!

adeas31 (approved):

Issue #460, opened 2026-04-01
ModelExecutionException: Timeout


psatke:
### Description

When i run a simulation model that takes more than 10 s, an exception TimeoutExpired is raised. I guess the initial definition in `ModelicaSystem.ModelExecutionCmd.timeout` is not updated by simargs={"alarm": 120}. I haven't found a solution to change the value of timeout besides overwriting the init value. 

### Steps to Reproduce

You can run the example.py file in the zip-archive, which will run a thermal building model created by the TEASER workflow. To run it you will need to install modelica library AixLib. The model should take longer than 10 s to calculate.

[Example.zip](https://github.com/user-attachments/files/26404348/Example.zip)

### Expected Behavior

When i specify a flag like `--alarm=120`, i expect the timeout value to be raised accordingly.

### Screenshots

<img width="1822" height="1056" alt="Image" src="https://github.com/user-attachments/assets/6885e505-9b69-40d2-ad9e-05d9dfc5b22a" />

### Version and OS

 - Python Version: 3.12.3
 - OMPython Version: dev https://github.com/OpenModelica/OMPython/commit/5b058215e550b42155382f16fb6edaea40617ed6
 - OpenModelica Version: 1.26.3~1-g7583224
 - OS: Linux Ubuntu 24.04.4 LTS

### Additional Context

I also had trouble using the `simulate(simargs={"override": {"stopTime": 63072000.0}})` feature. It did not raise an error but it also didn't change the time values in the result file, so it seems to be ignored. I don't know if the issues are related or if i made a mistake, since you've recently been working on the override feature (there is a typo in https://github.com/OpenModelica/OMPython/blob/5b058215e550b42155382f16fb6edaea40617ed6/OMPython/ModelicaSystem.py#L1345).

syntron:
OMC and OMPython use different timeouts - one within OMC and the other by Python using subprocess to call OMC

You should be able to modify the Python timeout as follows (main branch):

```
    omcs = OMPython.OMCSessionLocal(timeout=120.0)
    mod = OMPython.ModelicaSystemOMC(
        session=omcs,
    )
```

or using v4.0.1:

```
    omc = OMPython.OMCProcessLocal(timeout=120.0)
    mod = OMPython.ModelicaSystem(
         omc_process=omc,
    )
```

Untested - please check and provide feedback!

syntron:
`stopTime` is tested in `test_ModelicaSystem.test_getters()` and other unittests

[...]

PR #463, opened 2026-04-23
fix timeout handling

syntron:
the define timeout value (in OMCSession) is not propagated correctly to `ModelExecutionCmd()` - thus, any change does not have the desired effect

This PR additionally improves the log / error messages to also provide the value of the (OMPython) timeout.

One function (`set_timeout()`) was moved from OMCSessionABC to OMSessionABC

Reported in issue #460


syntron:
@adeas31 this could be added before PR #441

adeas31 (approved):

PR #444, opened 2026-02-27
F001 restructure cleanup

syntron:
(F001) cleanup after restructure
    
    [README.md] small updates
    
    [__init__] small updates

PR #469, opened 2026-06-27
OMParser update

syntron:
update OMParser; mainly just change code updates to get it safe for reuse in ModelicaDoE usecases

* please check as I do not have all details of the code workflow
* should be after PR #444 but it is save anytime as there is no other PR touching OMParser.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ OMPython is a Python interface that uses ZeroMQ to communicate with OpenModelica
 
 ## Dependencies
 
--   Python 3.x supported
--   PyZMQ is required
+  - Python >= 3.10 supported with complete functionality for Python >= 3.12
+  - Additional packages: numpy, psutil, pyparsing and pyzmq
 
 ## Installation
 
@@ -49,8 +49,8 @@ help(OMPython)
 ```
 
 ```python
-from OMPython import OMCSessionLocal
-omc = OMCSessionLocal()
+import OMPython
+omc = OMPython.OMCSessionLocal()
 omc.sendExpression("getVersion()")
 ```
 
--- a/OMPython/OMParser.py
+++ b/OMPython/OMParser.py
@@ -674,7 +674,7 @@ def get_the_set(parsed: OMParserData, string: str):
     next_set_list = []
     end = len(string)
 
-    if "{" and "}" in string:
+    if "{" in string and "}" in string:
         while index < len(string):
             character = string[index]
             if character == "{":
--- a/OMPython/OMParser.py
+++ b/OMPython/OMParser.py
@@ -658,8 +658,8 @@ def get_the_set(parsed: OMParserData, string: str):
     index = 0
     count = 0
     parsed.next_set[0] = ''
-    inner_sets = []
-    next_set_list = []
+    inner_sets: list[str] = []
+    next_set_list: list[str] = []
     end = len(string)
 
     if "{" in string and "}" in string:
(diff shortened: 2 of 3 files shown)


syntron:
@adeas31 The current code still uses Python2 definitions - thus it fails:
```
    def formatRecords(strings):
        result['RecordResults'] = {}
        recordName = strings[strings.find("record ") + 1:strings.find("\n")]
        recordName = recordName.replace("ecord ", '').strip()
        strings = strings.replace(("end " + recordName + ";"), '').strip()
        recordItems = strings[strings.find("\n") + 1: len(strings)]
>       recordItems = recordItems.translate(None, "\\")
E       TypeError: str.translate() takes exactly one argument (2 given)
```

Should it be converted to Python3 at all or removed? This PR tries to update the code to get it running (see: https://github.com/OpenModelica/OMPython/pull/469/commits/1b592c6e44f968b397b3b5b90d8b6f4a5386b5f1 for Python2 => Python3 conversion) and thread save but is it still be required or should it be completly removed? Overall it was not working for some time / any Python3 usage

adeas31:
As far as I can see the code is used to format the records, the record returned by `simulate` API is handled via `formatSimRes`. The other APIs that return a record are [getMessagesStringInternal](https://build.openmodelica.org/Documentation/OpenModelica.Scripting.getMessagesStringInternal.html) and [checkSettings](https://build.openmodelica.org/Documentation/OpenModelica.Scripting.checkSettings.html) or maybe more. You can test `getMessagesStringInternal` and see if its result record is formatted correctly.

I recommend converting the code to python 3 and adding a test for it.

syntron:
@adeas31 I'm not that familar with the OM return values to check this in detail; regarding OMParser, I noted that it it does not return the expected results:

running test_everything from test_OMParser (currently commented out; content of `results`)
```
{'RecordResults': {',5.9,6,NONE ( )},record AB': '"', 'startTime': 'ErrorLevel.warning', "'stop*Time'": 'SOME(1.0', 'nd ABC;}': 'end ABC;}', 'RecordName': ''}}
```

the same test in test_OMTypedParser:
```
(1.0, ((1, True, 3), ('4"\n', 5.9, 6, None), {'startTime': 'ErrorLevel.warning', "'stop*Time'": 1.0}))
```

It is the same input data to parse! The changes in this PR have no effect on `results` of `test_OMParser.test_everything()` - it is the same `results` data before / after applying it; just https://github.com/OpenModelica/OMPython/pull/469/commits/675997b44a4080d0e20fc0488186a934b9e6aa08 is always needed to fix Python 2 code.

Input data as reference:
```
    testdata = """
    (1.0,{{1,true,3},{"4\\"
",5.9,6,NONE ( )},record ABC
    startTime = ErrorLevel.warning,
    'stop*Time' = SOME(1.0)
end ABC;})
    """
```

README.md
```diff
@@ -8,8 +8,8 @@ OMPython is a Python interface that uses ZeroMQ to communicate with OpenModelica
 
 ## Dependencies
 
--   Python 3.x supported
--   PyZMQ is required
+  - Python >= 3.10 supported with complete functionality for Python >= 3.12
```
adeas31, line 11: Can't we simply say Python >= 3.10 or Python >= 3.12 supported. Do we need to be verbose here?

adeas31:
Can this be merged now or does it depend on any other PR?

README.md
syntron: we can go with Python >= 3.10; however, the OMPath functionality is only supported for Python >= 3.12 - Python 3.10 has only a compatibility layer to get something working

PS: this should be in the discussion for PR #444

syntron:
@adeas31 it can be merged as I hope to not change any functionality; however, I did only some basic tests as I have no real use cases for this part of the code ...

PR #444 should be run before - it is just some simple / small cleanup ...; else it is included in this PR ...

syntron:
I just run some checks; the return values from `OMParser` would crash OMPython ...

Example from run of `test_ModelicaSystemOMC.test_getSolutions()`:

```
print(expr) => buildModel(M, variableFilter=".*")
print(result) => {"/tmp/88ab0555-6bbe-4470-b43e-2d59df7e50d4/M", "M_init.xml"}
print(om_parser_typed) => ('/tmp/88ab0555-6bbe-4470-b43e-2d59df7e50d4/M', 'M_init.xml')
print(om_parser_basic) => {'SET1': {'Values': ['"/tmp/88ab0555-6bbe-4470-b43e-2d59df7e50d4/M", "M_init.xml"']}}
```

The data expected by the code is given by om_parser_typed; om_parser_basic would crash the code if returned; nevertheless, it contains the same / the expected information ...

Did I missed some additional conversion, i.e. SET1 => `set()`?

(test run with both versions of the code - same results!)

sjoelund:
If we do remove the untyped API from OM finally, the parsers aren't even needed. We could switch to `--interactiveDumpFormat=json`. Older OMC versions still need a parser though :(

PR #476, opened 2026-09-21
fix loadFile for package.mo: don't copy it out of its package directory

arun3688:
### Related Issues

https://github.com/OpenModelica/OpenModelica/issues/16771

### Purpose

load directory-based packages (package.mo) in place instead of copying just the one file,  it is possible the model might be located in sub-directories


adeas31:
@arun3688 would be nice to add a test about it so we don't break it in future.

arun3688:
@adeas31 sure i will add it

AnHeuermann:
I updated the Windows installer, let's see if the Windows nightly test passes now.

adeas31:
I think the Windows tests will still fail. See #478

